PR #30, opened 2015-04-27
Fix role/databag uploads for top-level dirs

jaymzh:


lib/taste_tester/config.rb
```diff
@@ -82,7 +82,11 @@ def self.databags
     end
 
     def self.relative_databag_dir
-      (base_dir && !base_dir.empty?) ? File.join(base_dir, databag_dir) : x
+      if base_dir && !base_dir.empty?
+        File.join(base_dir, databag_dir)
+      else
+        databagdir
```
odcinek: `databag_dir`?

jaymzh: fixed

PR #37, opened 2015-08-13
End-to-end tt test

odcinek:
This won't run on travis, as we're doing live ssh here. Best we can get is a local run.

### DIFF
--- a/test/run-git.sh
+++ b/test/run-git.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -e
+set -o verbose
+rm -rf /tmp/ops
+GIT="git -c user.email='foo@bar.com' -c user.name='foobar'"
+
+function tt {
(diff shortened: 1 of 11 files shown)


odcinek:
Updated, should be ok to merge

test/run-git.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -e
+set -o verbose
+rm -rf /tmp/ops
+GIT="git -c user.email='foo@bar.com' -c user.name='foobar'"
+
+function tt {
+bundle exec ./bin/taste-tester test -s localhost -y -c ./test/tt-git.rb -v
```
jaymzh: before committing, can you indent this line since it's the body of the function?

jaymzh: actually since that's the only thing i see here, I'll merge, but can you do  that in both tests when you do your next update, please?

PR #46, opened 2015-12-11
Do not exit from ssh error, raise instead

odcinek:


jaymzh:
the library doing this is fine, but please catch that elsewhere and exit nicely. no need to stacktrace on users, we already have a nice error here.

bin/taste-tester
```diff
@@ -362,28 +362,32 @@ MODES:
...
+    when :run
+      TasteTester::Commands.runchef
+    when :upload
+      TasteTester::Commands.upload
+    else
+      logger.error("Invalid mode: #{mode}")
+      puts parser
+      exit(1)
+    end
+  rescue
```
jaymzh: errr - that's... bringing the nuclear bomb to the knife fight. Raise a reasonable error and catch it. We shouldn't suppress exceptions that haven't printed nice errors for the user.

PR #48, opened 2016-03-22
updated Ohai plugin path config

patcox:
Fix this warning:

`WARN: Ohai::Config[:plugin_path] is set. Ohai::Config[:plugin_path] is deprecated and will be removed in future releases of ohai. Use ohai.plugin_path in your configuration file to configure :plugin_path for ohai.`

PR #49, opened 2016-03-23
Updated rubocop config, re-linted everything

patcox:
The rubocop config was out of date, as pointed out in  https://github.com/facebook/taste-tester/pull/48. I've updated it to configure the
most recent cops. I ran the new config across all existing files and
corrected things to get rid of the warnings and errors.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,12 +80,35 @@ FileName:
 PerceivedComplexity:
   Enabled: false
 
-TrailingComma:
-  Enabled: false
+TrailingCommaInLiteral:
+  EnforcedStyleForMultiline: comma
+
+TrailingCommaInArguments:
+  EnforcedStyleForMultiline: comma
 
 UnneededDisable:
   Enabled: false
 
+MultilineOperationIndentation:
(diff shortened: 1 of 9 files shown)


patcox:
grr rubocop

```
$ bundle exec rubocop
Inspecting 17 files
..CCC............
Offenses:
bin/taste-tester:118:28: C: Use !empty? instead of size > 0.
  mode = ARGV.shift unless ARGV.size > 0 && ARGV[0].start_with?('-')
                           ^^^^^^^^^^^^^
lib/taste_tester/client.rb:54:9: C: Always use raise to signal exceptions.
        fail "Could not open repo from #{TasteTester::Config.repo}"
        ^^^^
lib/taste_tester/commands.rb:86:9: C: Always use raise to signal exceptions.
        fail "Could not open repo from #{TasteTester::Config.repo}"
        ^^^^
17 files inspected, 3 offenses detected
```

I guess I need to config those cops too.

patcox:
How can I tell which version of rubocop is used in Travis? Can/should we pin it in `.travis.yml`?

PR #53, opened 2016-08-18
Add configurable chef config name

odcinek:


zfjagann:
:+1:

jaymzh:
I'm fine with thus, but we should follow up with `Path.join()`

Waiting for tests to pass.

PR #54, opened 2016-08-19
Add argument to specify the hostname of your local machine

natewalck:
This allows you to override @host in server.rb, which by default is Socket.gethostname. Socket.gethostname does not work reliably when DNS is not accurate, so this allows you to manually specify a hostname (IP addresses work as well).

I tested this using '[::1]' as well as my hostname and both situations worked as expected. Also tested setting my_hostname in taste_tester_config.rb and it worked as expected.

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -207,6 +207,12 @@ MODES:
     end
 
     opts.on(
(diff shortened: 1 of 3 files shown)


jaymzh:
:+1:

PR #56, opened 2016-09-02
Add locallink option - setup chef client for taste-tester on local sy…

joshuamiller01:
…stem

- Create a "locallink" option, which will operate upon the local chef client configuration without using ssh
- Set chef server port only when it has a value
- Do not touch repo when linkonly is set in test subcommand

### DIFF
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -87,7 +87,7 @@ module TasteTester
           raise "Could not open repo from #{TasteTester::Config.repo}"
         end
       end
-      unless TasteTester::Config.skip_pre_test_hook || 
+      unless TasteTester::Config.skip_pre_test_hook ||
           TasteTester::Config.linkonly
         TasteTester::Hooks.pre_test(TasteTester::Config.dryrun, repo, hosts)
       end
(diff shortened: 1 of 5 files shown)


lib/taste_tester/commands.rb
```diff
@@ -99,7 +101,7 @@ def self.test
           tested_hosts << hostname
         end
       end
-      unless TasteTester::Config.skip_post_test_hook
+      unless TasteTester::Config.skip_post_test_hook || TasteTester::Config.linkonly
```
jaymzh: I'm not convinced always skipping the hooks with -L makes sense... but I don't feel strongly.. but we don't know what people may put in their plugins.

jaymzh:
few minor comments, but in generaly I really like it.

lib/taste_tester/commands.rb
joshuamiller01: I've contemplated this but I think I want to leave it in.  My reasoning: our default hook functions want a repo, but I think the "spirit" of linkonly doesn't involve a repo (i.e. the help is "Only setup the remote server, skip uploading.")

PR #60, opened 2016-10-01
Fix class_eval invocation so that errors in hooks are reported properly.

jaymzh:


zfjagann:
Dup of #59

jaymzh:
Ok fucker, I'll let you take credit for my fix. =P

PR #63, opened 2016-11-01
Windows Native `taste-tester upload`

svmastersamurai:
These changes will let you `upload` a chef development repo to `taste-tester` natively on Windows.

* Windows cannot run `chef-zero` with the `--daemon` flag, so instead we create a scheduled task that will spawn a `chef-zero` process. The side-effect of this is that a window will remain open for the duration of the testing.
* Added a Windows module to contain windows specific code
* Shifted order of taste-tester issuing the `@state.wipe` when in `stop` mode.
* Added `pid` method to `State` class so it can infer the chef-zero process id
* Added `windows?` method to `Server` class so it only executes windows specific code.

### DIFF
--- a/lib/taste_tester/server.rb
+++ b/lib/taste_tester/server.rb
@@ -21,6 +21,7 @@ require 'timeout'
 require 'between_meals/util'
 require 'taste_tester/config'
 require 'taste_tester/state'
+require 'taste_tester/windows'
 
 module TasteTester
   # Stateless chef-zero server management
@@ -152,22 +153,31 @@ module TasteTester
                       :logging => TasteTester::Config.chef_zero_logging,
                     })
       logger.info("Starting chef-zero of port #{@state.port}")
-      cmd = "#{chef_zero_path} --host #{@addr} --port #{@state.port} -d"
-      if TasteTester::Config.chef_zero_logging
-        cmd << " --log-file #{@log_file} --log-level debug"
+      if windows?
+        extend ::TasteTester::Windows
+        start_win_chef_zero_server
+      else
+        cmd = "#{chef_zero_path} --host #{@addr} --port #{@state.port} -d"
+        if TasteTester::Config.chef_zero_logging
+          cmd << " --log-file #{@log_file}" +
+            ' --log-level debug'
+        end
+        cmd << ' --ssl' if TasteTester::Config.use_ssl
+        Mixlib::ShellOut.new(cmd).run_command.error!
       end
-      if TasteTester::Config.use_ssl
-        cmd << ' --ssl'
-      end
-      Mixlib::ShellOut.new(cmd).run_command.error!
     end
 
     def stop_chef_zero
-      logger.info('Killing your chef-zero instances')
-      s = Mixlib::ShellOut.new("pkill -9 -u #{ENV['USER']} -f bin/chef-zero")
-      s.run_command
-      # You have to give it a moment to stop or the stat fails
-      sleep(1)
+      if windows?
+        extend ::TasteTester::Windows
+        nuke_all_cz_pids
+      else
+        logger.info('Killing your chef-zero instances')
+        s = Mixlib::ShellOut.new("pkill -9 -u #{ENV['USER']} -f bin/chef-zero")
+        s.run_command
+        # You have to give it a moment to stop or the stat fails
+        sleep(1)
+      end
     end
 
     def chef_zero_path
@@ -183,6 +193,10 @@ module TasteTester
       end
       logger.error('chef-zero not found')
       exit(1)
+    end
+
+    def windows?
+      Gem.win_platform?
     end
   end
 end
--- a/lib/taste_tester/windows.rb
+++ b/lib/taste_tester/windows.rb
@@ -16,6 +16,10 @@
 
 module TasteTester
   module Windows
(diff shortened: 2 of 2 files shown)


lib/taste_tester/server.rb
```diff
@@ -84,14 +85,14 @@ def start
         logger.warn('Starting taste-tester server')
       end
       @state.wipe
-      write_config
       start_chef_zero
```
jaymzh: Out of curiosity why ?

lib/taste_tester/server.rb
```diff
@@ -152,19 +153,30 @@ def start_chef_zero
...
     def stop_chef_zero
-      logger.info('Killing your chef-zero instances')
-      s = Mixlib::ShellOut.new("pkill -9 -u #{ENV['USER']} -f bin/chef-zero")
-      s.run_command
-      # You have to give it a moment to stop or the stat fails
-      sleep(1)
+      if windows?
+        extend ::TasteTester::Windows
+        stop_win_chef_zero_server
+        @state.wipe
```
jaymzh: you do this in `stop` above, no need to do it here...

lib/taste_tester/windows.rb
```diff
@@ -0,0 +1,64 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module TasteTester
+  module Windows
+    PATH_TO_WRAPPER_SCRIPT =
+      File.absolute_path(File.join(
```
jaymzh: OK, so if I understand this all correctly, the only reason for the wrapper script is so that when the scheduled task gets run, we shove the PID into the statefile?

That's... terrible. I mean it's a neat hack... but Jesus, Microsoft.

scripts/win_chef_zero_server.ps1
```diff
@@ -0,0 +1,27 @@
...
+}
+
+$proc = Start-Process @chef_zero_server
+Start-Sleep 2
+
+# Steal the process ID and then inject it into the taste-tester state file
+# so it can be referenced in the plugin.
+$state = Get-Content ~/.chef/taste-tester-ref.json | ConvertFrom-JSON
+$state | Add-Member -Type NoteProperty -Name pid -Value $proc.id -Force
+$state |
```
jaymzh: There's a race-condition here... to be clear. but it's probably ... OK. We should at least add a comment to that effect though.

lib/taste_tester/windows.rb
svmastersamurai: Yeah the main problem is that Windows can't use the `--daemon` flag. There may be a better way to do this, but yeah I agree it is a big ball of poo 😢

scripts/win_chef_zero_server.ps1
svmastersamurai: Another way is to make the PID placement for all platforms. I was purposefully trying to keep the changes restricted to the bare minimum to get Windows operational but I can revise this to make it less hacky if need be!

lib/taste_tester/server.rb
svmastersamurai: 👍

svmastersamurai:
Alrighty, moved the `@state` mutilation into Ruby with the help of the `wmi-lite` gem. And uh... dealt with merging latest cut of `master`. I don't think I cherry-picked things right, but the feedback is making my additions look much cleaner! 🍨

jaymzh:
Oh god you merged somethign somewhere. Please rebase so the diffs are ontop of master :)

lib/taste_tester/windows.rb
```diff
@@ -0,0 +1,90 @@
...
+      )
+
+    def start_win_chef_zero_server
+      script_opts = ''
+      host = 'localhost'
+      port = @state.port
+      ps_opts = '-noprofile -noninteractive'
+      script_opts = '-UseSSL' if @state.ssl
+      cmd = format(
+        'powershell.exe %s -file %s -CZHost %s -CZPort %s %s',
```
jaymzh: now that we don't have the crazy pid thing, can't we just run chef-zero directly here? Do we need the wrapper script?

lib/taste_tester/server.rb
jaymzh: still curious why you switched the order here.

jaymzh:
Oh you'll probably have to force-push to your branch...

svmastersamurai:
Ah, i'll give it a try! Should get around to updates tomorrowish

svmastersamurai:
@jaymzh tbh I couldn't tell you _exactly why_ I changed the ordering aside from it was necessary at the time to get the first version working. I managed to refactor the code a bit and all those calls are now back to their original form.

This iteration removed all the powershell stuff, it looks like `Mixlib::ShellOut` was the culprit: it would just wait for the process to finish once it was created. If you're cool with using `Process.spawn` the end result is all the same, and with much less clowniness 😎 

Cleaned out my `git` log too, we should be 👍  there

jaymzh:
LGTM. I'll merge when tests pass.

jaymzh:
Thanks!

PR #64, opened 2016-11-01
Calculate time on remote end

odcinek:
In case of timezone (not time) difference between source and destination we can end up with invalid testing time. Touch command takes [[CC]YY]MMDDhhmm[.ss], which if calculated locally might end up being in the past on remote end (since there is no timezone information).

This fixes it by passing timestamp and calculating time on remote end.

### DIFF
--- a/lib/taste_tester/host.rb
+++ b/lib/taste_tester/host.rb
@@ -87,8 +87,7 @@ module TasteTester
       ssh = TasteTester::SSH.new(@name)
       ssh << 'logger -t taste-tester Moving server into taste-tester' +
         " for #{@user}"
-      ssh << "touch -t #{TasteTester::Config.testing_end_time}" +
-        " #{TasteTester::Config.timestamp_file}"
+      ssh << touchcmd
       ssh << "echo -n '#{@serialized_config}' | base64 --decode" +
         " > #{TasteTester::Config.chef_config_path}/client-taste-tester.rb"
       ssh << "rm -vf #{TasteTester::Config.chef_config_path}/" +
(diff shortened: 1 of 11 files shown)


lib/taste_tester/host.rb
```diff
@@ -188,21 +187,27 @@ def in_test?
...
-          " #{TasteTester::Config.timestamp_file}"
+        transport << touchcmd
         transport.run!
       end
     end
 
     private
 
+    def touchcmd
+      touch = Base64.encode64("touch --date \"$(date -d @" +
```
jaymzh: This doesn't work on macs:

```
[root@mactest989.06.ash3 ~]# date -d @1478049513 +'%Y-%m-%d %T'
usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]
```

jaymzh: Nor does your touch command:


```
[root@mactest989.06.ash3 ~]# touch --date
touch: illegal option -- -
usage:
touch [-A [-][[hh]mm]SS] [-acfhm] [-r file] [-t [[CC]YY]MMDDhhmm[.SS]] file ...
```

odcinek:
@jaymzh works on Darwin now

lib/taste_tester/host.rb
```diff
@@ -188,21 +187,30 @@ def in_test?
...
 
     private
 
+    def touchcmd
+      touch = Base64.encode64(
+        "if [ 'Darwin' = $(uname) ]; then touch -t \"$(date -r " +
+        "#{TasteTester::Config.testing_end_time.to_i} +'%Y%m%d%H%M.%S')\"" +
+        " #{TasteTester::Config.timestamp_file}; else touch --date \"$(date -d @" +
+        "#{TasteTester::Config.testing_end_time.to_i} +'%Y-%m-%d %T')\"" +
+        " #{TasteTester::Config.timestamp_file}; fi").delete("\n")
```
jaymzh: did you want `chomp` here?

odcinek: no, this needs to be contiguous string with no newlines, which Base64.encode64 sprinkles all over

PR #66, opened 2016-12-03
Just keep the port upon restart

### LINKED ISSUE #65: tt should try to keep the same port on restart
Otherwise, any in-flight test will fail.
It's really not hard to do, we just need to keep the port instead of wiping the entire config...

Babar:
Do not throw the baby with the bath water when shutting down.

closes #65

### DIFF
--- a/lib/taste_tester/config.rb
+++ b/lib/taste_tester/config.rb
@@ -95,8 +95,13 @@ module TasteTester
     end
 
     def self.chef_port
+      require 'taste_tester/state'
+      previous_ports = []
+      if TasteTester::State.port
+        previous_ports << TasteTester::State.port
+      end
       range = chef_port_range.first.to_i..chef_port_range.last.to_i
-      range.to_a.shuffle.each do |port|
+      (previous_ports + range.to_a.shuffle).each do |port|
         unless port_open?(port)
           return port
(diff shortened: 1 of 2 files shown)


lib/taste_tester/config.rb
```diff
@@ -95,8 +95,13 @@ def self.relative_databag_dir
...
 
     def self.chef_port
+      require 'taste_tester/state'
+      previous_ports = []
+      if TasteTester::State.port
+        previous_ports << TasteTester::State.port
+      end
       range = chef_port_range.first.to_i..chef_port_range.last.to_i
-      range.to_a.shuffle.each do |port|
+      (previous_ports + range.to_a.shuffle).each do |port|
```
jaymzh: it took me a minute to figure out why you made that an array. FWIW, I woulda done it this way:

```
range = (chef_port_range.first.to_i..chef_port_range.last.to_i).to_a
if TasteTester::State.port
  range.prepend(TasteTester::State.port)
end
range.shuffle.each do |port|
```

jaymzh:
Not waiting for tests since it's just space changes and the tests passed last time.

jaymzh:
Thanks @Babar !!

PR #71, opened 2017-02-04
Fix taste-tester tunneling commands.

flz:
This fixes two separate things:
- We try to run touch -t with an invalid parameter.
- The ServerAliveCountMax doesn't reflect the testing window.

### DIFF
--- a/lib/taste_tester/tunnel.rb
+++ b/lib/taste_tester/tunnel.rb
@@ -44,15 +44,13 @@ module TasteTester
     end
 
     def cmd
-      if TasteTester::Config.user != 'root'
-        pid = '$$'
-      else
-        pid = '\\$\\$'
-      end
-      cmds = "ps -p #{pid} -o pgid | grep -v PGID" +
+      @max_ping = @delta_secs / 10
+      pid = '$$'
+      @ts = TasteTester::Config.testing_end_time.strftime('%y%m%d%H%M.%S')
+      cmds = "ps -o pgid= -p $(ps -o ppid= -p #{pid}) | sed \"s| ||g\" " +
              " > #{TasteTester::Config.timestamp_file} &&" +
-             " touch -t #{TasteTester::Config.testing_end_time}" +
-             " #{TasteTester::Config.timestamp_file} && sleep #{@delta_secs}"
+             " touch -t #{@ts} #{TasteTester::Config.timestamp_file} &&" +
+             " sleep #{@delta_secs}"
       # As great as it would be to have ExitOnForwardFailure=yes,
       # we had multiple cases of tunnels dying
       # if -f and ExitOnForwardFailure are used together.
@@ -62,14 +60,14 @@ module TasteTester
       cmd = "#{TasteTester::Config.ssh_command} " +
             "-T -o BatchMode=yes -o ConnectTimeout=#{@timeout} " +
             '-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ' +
-            '-o ServerAliveInterval=10 -o ServerAliveCountMax=6 ' +
+            "-o ServerAliveInterval=10 -o ServerAliveCountMax=#{@max_ping} " +
             "-f -R #{@port}:localhost:#{@server.port} "
       if TasteTester::Config.user != 'root'
         cc = Base64.encode64(cmds).delete("\n")
         cmd += "#{TasteTester::Config.user}@#{@host} \"echo '#{cc}' | base64" +
                ' --decode | sudo bash -x"'
       else
-        cmd += "root@#{@host} \"#{cmds}\""
+        cmd += "root@#{@host} '#{cmds}'"
       end
       cmd
     end
@@ -85,7 +83,8 @@ module TasteTester
       end
       cmd = "( [ -s #{TasteTester::Config.timestamp_file} ]" +
             " && #{sudo}kill -9 -- " +
-            "-\$(cat #{TasteTester::Config.timestamp_file}); true )"
+            "-\$(cat #{TasteTester::Config.timestamp_file}) 2>/dev/null; " +
+            ' true )'
       ssh << cmd
       ssh.run!
     end


flz:
Test build is still failing due to unrelated lint errors.

bbbco:
Just got tunneling to work properly for our environment (Linux based, not OSX). I had a similar result for fixes, but figured I could tack my fixes on to this pull request via comments. Hope they are helpful!

lib/taste_tester/tunnel.rb
```diff
@@ -49,9 +49,10 @@ def cmd
       else
         pid = '\\$\\$'
       end
+      @max_ping = @delta_secs / 10
       cmds = "ps -p #{pid} -o pgid | grep -v PGID" +
              " > #{TasteTester::Config.timestamp_file} &&" +
-             " touch -t #{TasteTester::Config.testing_end_time}" +
+             " touch --date '#{TasteTester::Config.testing_end_time}'" +
```
bbbco: I used `" touch -t #{TasteTester::Config.testing_end_time.strftime('%y%m%d%H%M.%S')} " +` instead, as that is how it is constructed in `host.rb`

bbbco: I used `" touch -t #{TasteTester::Config.testing_end_time.strftime('%y%m%d%H%M.%S')} " +` instead, as that is how the timestamp  is constructed in `host.rb`

lib/taste_tester/tunnel.rb
```diff
@@ -49,9 +49,10 @@ def cmd
       else
         pid = '\\$\\$'
       end
+      @max_ping = @delta_secs / 10
       cmds = "ps -p #{pid} -o pgid | grep -v PGID" +
```
bbbco: The `ps -p #{pid} -o pgid` command never did retrieve the ssh process for me. Instead, this will retrieve the bash process that ran the command via ssh. Unfortunately, when it comes time to use this pid to kill the tunnel, it only kills the bash process and not the parent ssh tunnel. Instead of this, I used this line (courtesy of http://superuser.com/q/150117):
`cmds = "cat /proc/#{pid}/status | grep PPid | cut -f2 " +`

lib/taste_tester/tunnel.rb
jaymzh: You can't change this to --date - it's not the same and doesn't exist on OSX.

lib/taste_tester/tunnel.rb
jaymzh: PPID isn't the same thing as the progress group id. The goal is to get the PG id and then kill that. That requires a negation (kill -PGID)  ... perhaps we're missing that somewhere.

But you're probably right we should get the PGID of the parent...  so maybe:

```
ps -p $(cat /proc/#{pid}/status | grep PPid | cut -f2) -o pgid | pgrep -v PGID
```

lib/taste_tester/tunnel.rb
davide125: --date is not in SUS, but -d is and should be equivalent

flz:
Will test and update the diff in the next couple of days. Thanks for the review!

jaymzh:
@flz  I actually reviewed it the day you sent it... but somehow the "send review" button didn't hit and so comments were pending. I only noticed when I came to reply to the other comments. Sorry. :(

lib/taste_tester/tunnel.rb
bbbco: :+1:

flz:
I took inspiration from the suggestions in the comments and changed a couple more things:
- The weird case around pid is unclear to me but this is the only way I could get it to work on my side. Weird escaping with double quotes meant /etc/chef/test_timestamp wasn't being created properly.
- I used 'ps -o field=' instead of 'ps -o field | pgrep -v field'. This syntax works at least on Linux and MacOS. If this is a known compatibility issue, I'll revert since it's a cosmetic change.

lib/taste_tester/tunnel.rb
```diff
@@ -44,15 +44,13 @@ def run
     end
 
     def cmd
-      if TasteTester::Config.user != 'root'
-        pid = '$$'
-      else
-        pid = '\\$\\$'
```
jaymzh: this won't work when we re-quote you inside of `sudo "whatever"` so you need this back....

jaymzh:
the rest looks good.

plus failing tests.

lib/taste_tester/tunnel.rb
flz: It doesn't work with the backslashes though.

How do I test all the cases, I'm not running taste-tester as root. I'm not sure why it requires root permissions to begin with.

jaymzh: It requires root to write to /etc/chef, because chef runs as root... cause it starts system services and such? So if a non-root user can change where chef gets it's config than a non-root user can get anything... tt clearly needs root access on the other side.

jaymzh: Anyway maybe something else changed? You should check in with @mikedodge04 or @natewalck  - they use TT via sudo, so if they can confirm this doesn't break them, I'm happy with it.

flz: I understand the need for root access remotely. I don't understand why you'd need to use sudo locally though.

natewalck: This *should* be fine. taste-tester is always run as the user, but our tt-me script requires sudo to edit the config. tt-me doesn't wrap TT itself, it just does client.rb tweaking and symlinking.

jaymzh: @flz  it doesn't. It does `ssh 'sudo \"stuff here with $$\"'`

jaymzh: @natewalck  ... not on Linux when you're going directly, right? That's what I need you to test.

natewalck: @jaymzh Ah, fair point. I'm not as certain on the linux side. I'll see what I can find out.

natewalck: Ramesh tested this on our linux boxes. He said we should be good to go on this.

jaymzh: OK so I decided to take a look at this deeper.

This code doesn't seem to work for me... but *also* the existing code does not  work. Unmodified code with tt ssh'ing as root:

```
# cat test_timestamp
#
```

Damn. What about non-root?

```
# cat test_timestamp
#
```

Nope, same thing. what the... when did this break?

What about with @flz 's changes? Nope, still don't get a PID in that file. Hmmm, do the commands work?

```
$ ssh root@host 'ps -o pgid= -p $(ps -o ppid= -p $$)'
 457791
```

Yeah, that seems to work. Hmmm, what about with the file redirection?

```
$ ssh root@host 'ps -o pgid= -p $(ps -o ppid= -p $$) > /etc/chef/test_timestamp'
```

```
# cat test_timestamp
 458433
```

Hmm, so the commands seem roughly correct, but this doesn't actually work with or without root...

flz: Definitely worked for me while ssh'ing as root. That's odd.

jaymzh: You can test this internally on a devserver where you can ssh as a normal user or root. Swing by my desk if you want some help.

jaymzh: Ohhhh - the PID thing doesn't happen unless we'er using tunnels, so that's why...

but the current code works flawlessly. Let's see what happens when I use your code? Hmmmm, that works as well. So OK, this code seems to work, I'm OK with that... I'm curious though where you see it not working and in what way? Can you swing by so I can make sure we're not missing something?

jaymzh: @flz ^^

jaymzh: @flz  ping again.

flz:
Alright, I tested with tunnels as root and non-root and everything looks good.

The remaining rubocop errors ain't mine.

flz:
@jaymzh ping?

jaymzh:
This is good, but I have to fix the lint errors so I can merge without breaking the build, since you don't want to. :)

flz:
More like I have no idea what they are about. I looked some up and it seems like rubocop is linting the files as if they should be rails files.

jaymzh:
Yeah, the one rule about indentation needs to go away - I'll sink our rubopcy rules hopefully today...

jaymzh:
https://github.com/facebook/taste-tester/pull/74 cleans up the existing code.

jaymzh:
OK, rebase and repush @flz

jaymzh:
Thanks!

PR #72, opened 2017-02-07
add default windows paths

svmastersamurai:
The default paths to some of the settings isn't Windows friendly unless you explicitly provide it in a config file. Let's make this slightly more delightful! Not sure how we want to handle per-OS checks but this will at least get that discussion going 😉 

This works on my machine (sans `upload` since our repo structures are different) for `start` and `stop` commands without needing to tell `taste-tester` where everything is!

### DIFF
--- a/lib/taste_tester/config.rb
+++ b/lib/taste_tester/config.rb
@@ -56,6 +56,13 @@ module TasteTester
     chef_config 'client.rb'
     my_hostname nil
 
+    if Gem.win_platform?
+      timestamp_file 'C:/chef/test_timestamp'
+      config_file 'C:/chef/taste-tester-config.rb'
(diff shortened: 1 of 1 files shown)


lib/taste_tester/config.rb
```diff
@@ -56,6 +56,13 @@ module Config
     chef_config 'client.rb'
     my_hostname nil
 
+    if Gem.win_platform?
```
jaymzh: rather than overwrite these, I'd do:

```
# platform agnostic settings here
if Gem.win_platform?
  # windows-specific settings here
else
  # non-windows version of above here
end
```

jaymzh:
ping @svmastersamurai ?

PR #81, opened 2017-08-25
Add option to track_symlinks

mikedodge04:
Add option to track_symlinks to support change to Between Meals

https://github.com/facebook/between-meals/pull/74


jaymzh:
LGTM. Pending merging of BM diff.

PR #84, opened 2017-12-06
NoOp option for for TT

mikedodge04:
NoOp option for for TT

### DIFF
--- a/lib/taste_tester/host.rb
+++ b/lib/taste_tester/host.rb
@@ -21,6 +21,7 @@ require 'colorize'
 
 require 'taste_tester/ssh'
 require 'taste_tester/noop'
+require 'taste_tester/locallink'
 require 'taste_tester/tunnel'
 
 module TasteTester
@@ -72,7 +73,9 @@ module TasteTester
     end
 
     def get_transport
-      if TasteTester::Config.transport.casecmp('noop').zero?
+      if TasteTester::Config.transport.casecmp('locallink').zero?
+        transport = TasteTester::LocalLink.new
+      elsif TasteTester::Config.transport.casecmp('noop').zero?
         transport = TasteTester::NoOP.new
       else
         transport = TasteTester::SSH.new(@name)
--- a/lib/taste_tester/host.rb
+++ b/lib/taste_tester/host.rb
@@ -221,7 +221,7 @@ module TasteTester
         url << ":#{TasteTester::State.port}" if TasteTester::State.port
       end
       # rubocop:disable Metrics/LineLength
-      ttconfig = <<-eos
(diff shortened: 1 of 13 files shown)


lib/taste_tester/host.rb
```diff
@@ -72,8 +73,10 @@ def runchef
     end
 
     def get_transport
-      if TasteTester::Config.locallink
+      if TasteTester::Config.transport.casecmp('locallink').zero?
```
jaymzh: probably easier to do:

```
case TasteTester::Config.transport
when 'locallink'
  ...
when 'noop'
  ...
else
   ...
end
```

lib/taste_tester/noop.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    alias << add
+
+    def run
+      @status, @output = exec(cmd, logger)
+    end
+
+    def run!
+      @status, @output = exec!(cmd, logger)
```
jaymzh: The whole point of no-op is to NOT RUN anything on the remote host. Just don't do anything.

jaymzh:
This looks good to me. Did you test it at all?

mikedodge04:
No I will test in the morning.

lib/taste_tester/noop.rb
```diff
@@ -0,0 +1,68 @@
...
+      [0, "# TasteTester by #{@user}"]
+    end
+
+    private
+
+    def cmd
+      @cmds.each do |cmd|
+        logger.info("No-op, faking run of: '#{cmd}' on #{@host}")
+      end
+      cmds = @cmds.join(' && ')
```
jaymzh: You can drop the rest of this. It doesn't matter, just as long as we have a track record of what we're not doing in debug mode.

lib/taste_tester/host.rb
```diff
@@ -71,13 +72,22 @@ def runchef
       end
     end
 
+    def print_noop_warning
+      @PRINTED_WARNING ||= logger.warn(
```
jaymzh: You can do that, but it's easier to do it the initializer for NoOp, and use a class variable (@@)

lib/taste_tester/noop.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    def print_noop_warning
+      @@PRINTED_WARNING ||= logger.warn(
+        'No-op plugin active, no remote commands will be run!'
+      )
+    end
+
+    def run!
+      print_noop_warning
```
jaymzh: I'd move this to the initializer. And add comment on why it's a class var and not an instance var

PR #88, opened 2018-04-09
Return failure and partial failure on test

malmond77:
Taste-testing on a server that is already configured by someone else, or has some other problem does not surface to the user except as a warning. This adds two new return codes which can be used in external automation.


lib/taste_tester/commands.rb
```diff
@@ -106,6 +106,21 @@ def self.test
...
+      # because we're building the list in the same order
+      if tested_hosts == hosts
+        # no exceptions, complete success
+        exit(0)
+      end
+      if tested_hosts.empty?
+        # nothing worked, complete failure
+        exit(3)
+      end
+      # otherwise, we got a mix of success and failure. This isn't well
```
jaymzh: This is highly misleading.

These are ones we *chose* not to test because they were already in testing.

That's very different from failed to test. If we fail to test any host, we immediately crash/bail out (if the host is down, if you don't have ssh access, etc.)

I'm OK with a different exit code here... but both the comments here and the documentation you added above is very misleading.

jaymzh:
also. your tests are failing. :)

lib/taste_tester/commands.rb
malmond77: If you taste-test one or more hosts, and the end result is that the taste-testing state of the selected hosts does not change then the test command did not succeed.

host.in_test? is specifically asserting the user is the same. If someone else is testing, then you are not testing.

Other failure modes like inability to connect are also fatal, but there are no changes to this.

malmond77:
On the tests: Fixing the rubocop errors is trivial, but it breaks in prod: Example:

```
[user@server ~/local/taste-tester] /opt/chef/embedded/bin/rubocop
Inspecting 20 files
....................

20 files inspected, no offenses detected
[user@server ~/local/taste-tester] /opt/chef/embedded/bin/rubocop -v
0.52.0
```

If I change the `.rubocop.yml` I get:

```
[user@server ~/local/taste-tester] /opt/chef/embedded/bin/rubocop
Warning: unrecognized cop TrailingCommaInArrayLiteral found in .rubocop.yml
Warning: unrecognized cop TrailingCommaInHashLiteral found in .rubocop.yml
Inspecting 20 files
..CC...C...C....C...

Offenses:

bin/taste-tester:251:19: C: Style/TrailingCommaInLiteral: Avoid comma after the last item of a hash.
          :m => 60,
                  ^
lib/taste_tester/client.rb:139:53: C: Style/TrailingCommaInLiteral: Avoid comma after the last item of a hash.
            TasteTester::Config.relative_databag_dir,
                                                    ^
lib/taste_tester/host.rb:145:64: C: Style/TrailingCommaInLiteral: Avoid comma after the last item of an array.
        'logger -t taste-tester Returning server to production',
                                                               ^
lib/taste_tester/server.rb:154:72: C: Style/TrailingCommaInLiteral: Avoid comma after the last item of a hash.
                      :logging => TasteTester::Config.chef_zero_logging,
                                                                       ^
lib/taste_tester/commands.rb:171:27: C: Style/TrailingCommaInLiteral: Avoid comma after the last item of an array.
        'Object not found',
                          ^

20 files inspected, 5 offenses detected
```

I'd be happy to fix this, but I think this is a different, and unrelated pull request.

jaymzh:
You're misunderstanding my comments.

According to your new diff you say "if we had partial success we have this new exit code" - but if we fail because the hostname is bad or you don't have access, then we actually **bail out**. Therefore your documentation is in fact **lying** in many cases.

The **only** case in which your new exit codes will be returned is when we chose not to test on a subset (or all) hosts.

Your documentation implies that if I do `taste-tester test -ys fakehost,realhost` that I'll get a return value of 2... but I won't. I will get a stacktrace (by design).

jaymzh:
This looks good to me. You just need fix the .rubocop.yml that's in this repo.

PR #92, opened 2018-09-06
taste-tester test race condition fixes

malmond77:
Taste-tester was originally conceived as a tool for humans. Adding automation has highlighted issues when the same host is accidentally chosen. This already gives an error, explaining that someone else is testing

> Taste-testing on _hostname_
> ERROR: User someoneelse is already testing on _hostname_
> ERROR: Taste-testing failed. HOST NOT PUT IN TESTING MODE.

The existing code uses the transport (usually SSH) multiple times which is slow, and therefore more likely to hit a race condition. Current code:

1. Checks for timestamp file presence
2. greps for username in config
3. greps for username in config file to see if it matches the current user
4. Writes the new config file out

It could be up to 6 connections, if the legacy FB stuff for matching usernames was being used.

Instead of using transport 4-6 times, we use it once, and pile a bunch of sequential logic in to predicate the addition of the new config file only if the host isn't being tested by someone else. Once the files are
written, symlinks changed, we look again to see if we "won".

Additional items:

* Removed timestamp check when reading the config file for an existing user. If the current config file mentions a user, it will definitely still be taste-tested. This check didn't help anything.
* `chef_config_path` is now written to a temporary file then renamed which makes the contents atomic. Two simultaneous writes can produce garbage configuration.
* Removed most explicit `rm -f` before `ln -s`, replacing with `ln -sf` which according to POSIX specifications also unlinks but in the same command.
* Changed single quote/apostrophe from use of `serialized_config` to double quotes. The command is already run with single quotes via SSH so this was effectively *un*quoting the text, which is more dangerous than double quoting or leaving unquoted.
* `@serialized_config` is never used as a member variable, so demoted to a local instance variable
* Removed FB legacy username lookup from `who_is_testing`.
* Move test config filename into constant.
* Use constant for preable written, and matched.
* Whole stack of rubocop config changes, and updates to make the tests pass.

### DIFF
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # vim: syntax=ruby:expandtab:shiftwidth=2:softtabstop=2:tabstop=2
 
 # Copyright 2013-present Facebook
@@ -66,7 +68,7 @@ module TasteTester
       unless TasteTester::Config.yes
         printf("Set #{TasteTester::Config.servers} to test mode? [y/N] ")
         ans = STDIN.gets.chomp
-        exit(1) unless ans =~ /^[yY](es)?$/
+        exit(1) unless ans.match?(/^[yY](es)?$/)
       end
       if TasteTester::Config.linkonly && TasteTester::Config.really
         logger.warn('Skipping upload at user request... potentially dangerous!')
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -19,6 +19,7 @@ require 'taste_tester/host'
 require 'taste_tester/config'
 require 'taste_tester/client'
 require 'taste_tester/logging'
+require 'taste_tester/exceptions'
 
 module TasteTester
   # Functionality dispatch
@@ -93,12 +94,11 @@ module TasteTester
       tested_hosts = []
       hosts.each do |hostname|
         host = TasteTester::Host.new(hostname, server)
-        if host.in_test?
-          username = host.who_is_testing
-          logger.error("User #{username} is already testing on #{hostname}")
-        else
+        begin
           host.test
           tested_hosts << hostname
+        rescue TasteTester::Exceptions::AlreadyTestingError => e
+          logger.error("User #{e.username} is already testing on #{hostname}")
         end
       end
       unless TasteTester::Config.skip_post_test_hook ||
--- a/lib/taste_tester/host.rb
+++ b/lib/taste_tester/host.rb
@@ -146,7 +146,7 @@ module TasteTester
         @name,
       )
 
-      if cmds && cmds.any?
+      if cmds&.any?
         transport = get_transport
         cmds.each { |c| transport << c }
         transport.run!
@@ -185,14 +185,14 @@ module TasteTester
       #    short circuits the test verb
       # This is written as a squiggly heredoc so the indentation of the awk is
       # preserved. Later we remove the newlines to make it a bit easier to read.
-      shellcode = <<~EOF
+      shellcode = <<~ENDOFSHELLCODE
         awk "\\$0 ~ /^#{USER_PREAMBLE}/{
           if (\\$NF != \\"#{@user}\\"){
             print \\$NF;
             exit 42
           }
         }" #{config_file}
-      EOF
+      ENDOFSHELLCODE
       shellcode.delete("\n")
     end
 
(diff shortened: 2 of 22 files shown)


lib/taste_tester/host.rb
```diff
@@ -23,12 +23,16 @@
...
 require 'taste_tester/tunnel'
+require 'taste_tester/exceptions'
 
 module TasteTester
   # Manage state of the remote node
   class Host
     include TasteTester::Logging
 
+    TASTE_TESTER_CONFIG = 'client-taste-tester.rb'
+    USER_PREABLE = '# TasteTester by '
```
joshuamiller01: PREAMBLE?

lib/taste_tester/host.rb
```diff
@@ -220,11 +228,11 @@ def config
       if TasteTester::Config.use_ssh_tunnels
         url = "#{scheme}://localhost:#{@tunnel.port}"
       else
-        url = "#{scheme}://#{@server.host}"
+        url = +"#{scheme}://#{@server.host}"
```
jaymzh: this is ... bad syntax afact?

lib/taste_tester/server.rb
```diff
@@ -158,7 +160,7 @@ def start_chef_zero
         extend ::TasteTester::Windows
         start_win_chef_zero_server
       else
-        cmd = "#{chef_zero_path} --host #{@addr} --port #{@state.port} -d"
+        cmd = +"#{chef_zero_path} --host #{@addr} --port #{@state.port} -d"
```
jaymzh: what is this leading plus thing?

taste_tester.gemspec
```diff
@@ -1,7 +1,9 @@
+# frozen_string_literal: true
+
 # rubocop:disable Metrics/BlockLength
 Gem::Specification.new do |s|
   s.name = 'taste_tester'
-  s.version = '0.0.13'
+  s.version = '0.0.14'
```
jaymzh: please do not do version bumps with other changes.

jaymzh:
This is really nice, thanks Matthew. But we definitely don't want to have a version bump at the same time. Lets get the features in, when we want to do a release we can. We'll want to build and deploy internally and make sure it works, then I can bump and build a gem.

lib/taste_tester/server.rb
malmond77: Marks a string as mutable (by making it an expression) when the default is frozen

lib/taste_tester/host.rb
```diff
@@ -23,12 +25,16 @@
...
 require 'taste_tester/tunnel'
+require 'taste_tester/exceptions'
 
 module TasteTester
   # Manage state of the remote node
   class Host
     include TasteTester::Logging
 
+    TASTE_TESTER_CONFIG = 'client-taste-tester.rb'
+    USER_PREAMBLE = '# TasteTester by '
```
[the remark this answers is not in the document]
malmond77: This is exactly the same string as used today: https://github.com/facebook/taste-tester/blob/master/lib/taste_tester/host.rb#L227

I agree this is not grammatically sound. Changing it would involve a backward compatibility hack else we'd lose the mutual exclusion aspect entirely. I opt to be laz^Wefficient here and focus on the main race condition problems.

lib/taste_tester/commands.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
malmond77: This codebase is relatively small. I've already added all these magic comments, and fixed up the few places where we assumed mutable strings. I'd like to keep this so we learn the cost of being Ruby 3 compatible in this regard.

lib/taste_tester/host.rb
```diff
@@ -94,22 +100,45 @@ def test
...
         " for #{@user}"
       transport << touchcmd
-      transport << "/bin/echo -n '#{@serialized_config}' | base64 --decode" +
-        " > #{TasteTester::Config.chef_config_path}/client-taste-tester.rb"
-      transport << "rm -vf #{TasteTester::Config.chef_config_path}/" +
-        TasteTester::Config.chef_config
-      transport << "( ln -vs #{TasteTester::Config.chef_config_path}" +
-        "/client-taste-tester.rb #{TasteTester::Config.chef_config_path}/" +
+      # shell redirection is also racey, so make a temporary file first
+      transport << "tt=$(mktemp #{TasteTester::Config.chef_config_path}/" +
```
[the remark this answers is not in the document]
malmond77: Agreed

lib/taste_tester/host.rb
malmond77: Nope, mutable string: https://stackoverflow.com/a/37799399

taste_tester.gemspec
malmond77: I'll back this bump out.

lib/taste_tester/host.rb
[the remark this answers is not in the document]
jaymzh: Yup, agreed. Just mis-remembered.

lib/taste_tester/host.rb
jaymzh: Oh jesus. This seems... onerous to me. but w/e

malmond77:
@jaymzh : the tests pass, and all your comments are resolved, with the notable exception of using frozen strings. It does work here, and I'd like to keep it.

jaymzh:
I dislike the frozen string stuff, but it's minimal invasion on this code base for now, so ... off we go!

Issue #82, opened 2017-10-23
Plugin Example


clburlison:
Is there any chance someone can upload a plugin sample? This repo is just different enough from chefctl that I'm configuring something incorrectly.

This is what I'm attempting to add as a plugin.
```ruby
module TasteTesterCustom
    def test_remote_client_rb_extra_code(_hostname)
      # Append to client.rb
      return "json_attribs '/etc/chef/run-list.json'"
  end
end

TasteTester::Hooks.register TasteTesterCustom
```

Thanks!

clburlison:
I ended up figuring it out.

```ruby
Hooks.class_eval do
  def self.test_remote_client_rb_extra_code(_hostname)
    %(
      json_attribs '/etc/chef/run-list.json'
      node_name '#{serial}'
      )
  end
end
```

PR #95, opened 2018-09-27
README fix, plugin_path does not have a default

bwann:
* In the README, it says plugin_path has a default to /etc/taste-tester-plugin.rb :

`plugin_path (string, default: /etc/taste-tester-plugin.rb)`

but in code it's really set to`nil`: https://github.com/facebook/taste-tester/blob/master/lib/taste_tester/config.rb#L40

This PR updates README to say the command line flag has no default, but a suggested path is added in the plugin documentation.

(or should the code use a default plugin path instead?)

* Add a short example plugin (see also https://github.com/facebook/taste-tester/issues/82) to README
* Markdown fixes to add credibility to claims I contributed to an open-source project ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ See the help for futher information.
 
 ## Prerequisites
 
-* Taste Tester assumes that /etc/chef/client.rb and /etc/chef/client.pem on your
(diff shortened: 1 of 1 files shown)


jaymzh:
Thanks!

PR #96, opened 2018-10-11
don't log repo information if --skip-repo-checks is supplied

dkaplan1:


jaymzh:
What's the motivation here? TT requires a repo, and logging info about it is wholly independent of whether you run site-specific sanity checks on it.

dkaplan1:
After speaking to Phil, going to close this and rework it.

PR #102, opened 2018-10-16
Add status accessor to noop, locallink modes

malmond77:
Before:

```
/opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/lib/taste_tester/host.rb:130:in `test': undefined method `status' for #<TasteTester::NoOp:0x0000000000c3aac8> (NoMethodError)
        from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/lib/taste_tester/commands.rb:99:in `block in test'
        from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/lib/taste_tester/commands.rb:96:in `each'
        from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/lib/taste_tester/commands.rb:96:in `test'
        from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/bin/taste-tester:400:in `<module:TasteTester>'
        from /opt/chefdk/embedded/lib/ruby/gems/2.4.0/gems/taste_tester-0.0.13/bin/taste-tester:32:in `<top (required)>'
        from /opt/chefdk/embedded/bin/taste-tester:23:in `load'
        from /opt/chefdk/embedded/bin/taste-tester:23:in `<main>'
```

After:

*no error*

### DIFF
--- a/lib/taste_tester/noop.rb
+++ b/lib/taste_tester/noop.rb
@@ -22,7 +22,7 @@ module TasteTester
     include TasteTester::Logging
     include BetweenMeals::Util
 
-    attr_reader :output
+    attr_reader :output, :status
 
     def initialize
       print_noop_warning
@@ -48,7 +48,7 @@ module TasteTester
     alias << add
 
     def run
-      run!
+      @status, @output = run!
     end
 
     def run!
--- a/lib/taste_tester/locallink.rb
+++ b/lib/taste_tester/locallink.rb
@@ -22,7 +22,7 @@ module TasteTester
     include TasteTester::Logging
     include BetweenMeals::Util
 
-    attr_reader :output
+    attr_reader :output, :status
 
     def initialize
       @host = 'localhost'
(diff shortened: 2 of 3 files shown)


lib/taste_tester/noop.rb
```diff
@@ -22,7 +22,7 @@ class NoOp
     include TasteTester::Logging
     include BetweenMeals::Util
 
-    attr_reader :output
+    attr_reader :output, :status
```
dkaplan1: Should we also initialize `@status = 0` so we don't run error() unnecessarily?

jaymzh: Line 56 always returns 0

jaymzh:
Thanks!

lib/taste_tester/noop.rb
malmond77: Looking through the code, these members aren't used much. They're only valid if transport.run is called first so I've refactored the code to return these directly.

PR #103, opened 2018-10-16
Add status addr to TasteTester::NoOp

dkaplan1:
TasteTester::Host::test expects a transport status attr (https://github.com/facebook/taste-tester/blob/master/lib/taste_tester/host.rb#L130) which NoOp does not currently have.


dkaplan1:
didn't see malmond77 already got to it :)

PR #104, opened 2018-10-17
Use return value of transport.run instead of status, output members

malmond77:
Was intended in PR 102, but it got merged before I could fix it.

### DIFF
--- a/lib/taste_tester/locallink.rb
+++ b/lib/taste_tester/locallink.rb
@@ -22,8 +22,6 @@ module TasteTester
     include TasteTester::Logging
     include BetweenMeals::Util
 
-    attr_reader :output, :status
-
     def initialize
       @host = 'localhost'
       @cmds = []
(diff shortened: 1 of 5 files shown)


lib/taste_tester/locallink.rb
```diff
@@ -36,11 +34,11 @@ def add(string)
     alias << add
 
     def run
-      @status, @output = exec(cmd, logger)
+      return exec(cmd, logger)
```
jaymzh: return is unnecessary

lib/taste_tester/locallink.rb
```diff
@@ -36,11 +34,11 @@ def add(string)
...
     end
 
     def run!
-      @status, @output = exec!(cmd, logger)
+      return exec!(cmd, logger)
```
jaymzh: return is unnecessary

lib/taste_tester/noop.rb
```diff
@@ -48,7 +46,7 @@ def add(string)
     alias << add
 
     def run
-      @status, @output = run!
+      return run!
```
jaymzh: return is unnecessary

PR #109, opened 2019-03-10
Support overriding SSH ConnectTimeout through config

omribahumi:
ssh.rb/tunnel.rb have hard coded 5 seconds as ConnectTimeout

Move this value to config

Tested this by replacing ssh command (`-e`) with something that logs ssh executions, seen `-o ConnectTimeout` change when changing the `--ssh-connect-timeout` flag; Also, saw the previous default (5) when not specifying the `--ssh-connect-timeout` option

### DIFF
--- a/lib/taste_tester/ssh.rb
+++ b/lib/taste_tester/ssh.rb
@@ -24,9 +24,8 @@ module TasteTester
 
     attr_reader :output, :status
 
-    def initialize(host, timeout = 5, tunnel = false)
+    def initialize(host, tunnel = false)
       @host = host
-      @timeout = timeout
       @tunnel = tunnel
       @cmds = []
     end
@@ -68,7 +67,8 @@ and come back once that works
       end
       cmds = @cmds.join(' && ')
       cmd = "#{TasteTester::Config.ssh_command} " +
-            "-T -o BatchMode=yes -o ConnectTimeout=#{@timeout} " +
+            '-T -o BatchMode=yes ' +
+            "-o ConnectTimeout=#{TasteTester::Config.ssh_connect_timeout} " +
             "#{TasteTester::Config.user}@#{@host} "
       if TasteTester::Config.user != 'root'
         cc = Base64.encode64(cmds).delete("\n")
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -328,6 +328,12 @@ MODES:
       options[:ssh_command] = c
     end
 
+    opts.on(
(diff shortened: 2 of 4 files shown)


lib/taste_tester/ssh.rb
```diff
@@ -24,9 +24,8 @@ class SSH
 
     attr_reader :output, :status
 
-    def initialize(host, timeout = 5, tunnel = false)
+    def initialize(host, tunnel = false)
```
jaymzh: this would break any user of `tunnel`, so I went searching around... and it turns out, it's not used. Not by this class, and not by any callers, we should just nuke it. AFAICT, this was designed to be used by tunnel.rb, but it doesn't use it, at all.

but since you would be breaking it, lets kill it in this PR

lib/taste_tester/ssh.rb
```diff
@@ -68,7 +67,7 @@ def cmd
       end
       cmds = @cmds.join(' && ')
       cmd = "#{TasteTester::Config.ssh_command} " +
-            "-T -o BatchMode=yes -o ConnectTimeout=#{@timeout} " +
+            '-T -o BatchMode=yes ' +
```
jaymzh: we should still pass in a connecttimeout, even if its the one from the config?

omribahumi: You can override this things by changing your ssh_config (either the one in `/etc/ssh` or the one in `~/.ssh`), or replace the SSH command with something that injects this parameter.

There is no default in the ssh_config man page, so I assume this will just the OS default TCP connect timeout?

I think we shouldn't specify this at all. Up to you.

jaymzh: It's really really long... we set this to 5 seconds by default, IIRC, becasue otherwise when you tt on a bunch of hosts, if one is down, it waits for that one forever.... honestly, I've considered changing it to 2...

joshuamiller01: I resolve this conversation.

jaymzh:
I agree this is fine. However, is it necessary if we finish up https://github.com/facebook/taste-tester/pull/110 ?

omribahumi:
@jaymzh how does #110 affect this one?
#110 is for keeping the tunnel up while the canary is still running, this one is for when it takes a long time to connect to a host (more than 5 seconds in the hard coded value)

jaymzh:
@omribahumi oh, duh, you're totally right. Ignore my comment. :)

PR #110, opened 2019-03-10
SSH tunnel will share the faith of timestamp file

omribahumi:
Instead of static duration, watch the timestamp file.

I'm not observing mtime here, as something may extend the canary duration between mtime expiration and next chef run (which I assume will delete the file?)

Tested this by:
1. running taste-tester
2. seeing my `sleep 60` in ps
3. deleting the timestamp file
4. `sleep 60` gone from ps

### DIFF
--- a/lib/taste_tester/tunnel.rb
+++ b/lib/taste_tester/tunnel.rb
@@ -26,12 +26,6 @@ module TasteTester
       @host = host
       @server = server
       @timeout = timeout
-      if TasteTester::Config.testing_until
-        @delta_secs = TasteTester::Config.testing_until.strftime('%s').to_i -
-                      Time.now.strftime('%s').to_i
-      else
-        @delta_secs = TasteTester::Config.testing_time
-      end
     end
 
     def run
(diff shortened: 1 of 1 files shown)


lib/taste_tester/tunnel.rb
```diff
@@ -44,13 +38,17 @@ def run
...
       @ts = TasteTester::Config.testing_end_time.strftime('%y%m%d%H%M.%S')
+
+      # Tie the faith of our SSH tunnel with the faith of timestamp file.
+      # taste-testing can be renewed, so we'll wait until the timestamp file
+      # is entirely gone before we kill the tunnel
       cmds = "ps -o pgid= -p $(ps -o ppid= -p #{pid}) | sed \"s| ||g\" " +
              " > #{TasteTester::Config.timestamp_file} &&" +
              " touch -t #{@ts} #{TasteTester::Config.timestamp_file} &&" +
-             " sleep #{@delta_secs}"
+             "while [ -f #{TasteTester::Config.timestamp_file} ]; do " +
```
jaymzh: this is race-y. If we try to tt again within those 60 seconds (which is likely if someone is taking over testing, they'll untest/test in rapid succession), this will break.

I wonder if we want to put our own PID into the timestamp file and have taste-untester updated to kill both?

omribahumi: I'm less concerned about the tunnel outliving the test, I'm worried about the test outliving the tunnel.

In current implementation, Chef will fail. In the case you're describing, we may possibly have stale tunnels, which is not a big concern IMHO. Am I missing something here?

Since this is still better than the existing implementation, can we land this and create a followup?
As for your fix suggestion, I don't fully understand the lifecycle of a test. For example, is there just a single code path that deletes the timestamp file?

jaymzh: Yes there is a single codepath that deletes the timestamp file - see taste-untester in this repo.

I think trading one bug for another isn't a good idea - we have non-Facebook users who may have worked around this bug, but aren't expecting long-lived stale tunnels (which is a potential security issue) on hosts.

If we put both PIDs in the timestamp file and then update taste-untester to do the right thing, I think that solves both problems.

jaymzh:
ping?

lib/taste_tester/tunnel.rb
joshuamiller01: I don't follow the discussion here.  It looks like taste-untester already kills the PGID, which we already put in the file; is killing the PGID somehow not enough?

jaymzh: Interesting point. So we take the pid of the shell we ssh'd in, get ITS pgid.

While that file is still there we sleep. In theory, the shell that's running the `while... sleep` is in the same process group, so in THEORY, if we kill pgid we should kill this shell and the tunnel. If the shell dies, that should be enough. Sure, I'm OK with that, if it works.

BTW, this has been modified since I last looked at this, I'm not sure what this line is supposed to be testing:

```
             "\"$(ps -o pgid= -p $(ps -o ppid= -p #{pid}) | sed \"s| ||g\")\" == \"$SSH_PGID\" ]]; do " +
```

making sure we didn't get re-parented is the idea there? I think we can keep our same pgid even if we get re-parented?

jaymzh:
Superceded by #133

PR #111, opened 2019-03-26
Include error message when tunnel setup fails

joshuamiller01:
Include error message when tunnel setup fails in order to ease troubleshooting.

### DIFF
--- a/lib/taste_tester/tunnel.rb
+++ b/lib/taste_tester/tunnel.rb
@@ -38,8 +38,8 @@ module TasteTester
       @port = TasteTester::Config.tunnel_port
       logger.info("Setting up tunnel on port #{@port}")
       @status, @output = exec!(cmd, logger)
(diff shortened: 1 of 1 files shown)


lib/taste_tester/tunnel.rb
```diff
@@ -38,8 +38,8 @@ def run
       @port = TasteTester::Config.tunnel_port
       logger.info("Setting up tunnel on port #{@port}")
       @status, @output = exec!(cmd, logger)
-    rescue StandardError
-      logger.error 'Failed bringing up ssh tunnel'
+    rescue StandardError => e
+      logger.error "Failed bringing up ssh tunnel: #{e}"
```
jaymzh: `e.message` is a bit better...

joshuamiller01: Tried both just now; seems equivalent.

PR #112, opened 2019-06-04
Add impact mode

emroch:
Create a new subcommand for measuring the impact of cookbook/role changes.

`taste-tester impact` will calculate the dependencies of the roles in the Chef repository and compare with modified files in the local working copy.  Potentially affected roles are reported and passed to an optional post_impact hook for further processing.

Note: version number bumped to 0.14

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -343,6 +343,13 @@ MODES:
       options[:yes] = true
     end
 
+    opts.on(
+      '--json', 'Format output as JSON for programatic consumption.' +
+      ' Default to false. Works on impact.'
+    ) do
+      options[:json] = true
+    end
+
     opts.separator ''
     opts.separator 'Control local hook behavior with these options:'
 
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -375,6 +375,18 @@ MODES:
     ) do
       options[:skip_post_test_hook] = true
     end
+
+    opts.on(
+      '--use-custom-impact-hook', 'Use customized impact hook. Works on impact.'
+    ) do
+      options[:use_custom_impact_hook] = true
+    end
+
+    opts.on(
+      '--skip-post-impact-hook', 'Skip post-impact hook. Works on impact.'
+    ) do
+      options[:skip_post_impact_hook] = true
+    end
   end
   # rubocop:enable Metrics/BlockLength
 
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -377,12 +377,6 @@ MODES:
     end
 
     opts.on(
-      '--use-custom-changeset-hook', 'Find a custom changeset. Works on impact.'
-    ) do
-      options[:use_custom_changeset_hook] = true
-    end
-
-    opts.on(
       '--use-custom-impact-hook', 'Use customized impact hook. Works on impact.'
     ) do
       options[:use_custom_impact_hook] = true
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -379,7 +379,7 @@ MODES:
     opts.on(
       '--use-custom-changeset-hook', 'Find a custom changeset. Works on impact.'
     ) do
-        options[:use_custom_changeset_hook] = true
+      options[:use_custom_changeset_hook] = true
     end
 
     opts.on(
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -90,6 +90,13 @@ MODES:
     cookbooks and roles. It also does a fair amount of sanity checking on
     your repo and you may specify --skip-repo-checks to bypass this.
 
+  profile
+    Examine local repo changes to determine which roles are potentially
+    affected by the changes. Compares the set of modified cookbooks with the
+    dependency lists of each role and reports any role which depends on a
+    modified cookbook. It is recommended to test your changes on at least one
+    server of each role, potentially also on multiple platforms.
+
   keeptesting
     Extend the testing time on server specified by -s by 1 hour unless
     otherwise specified by -t.
@@ -414,6 +421,8 @@ MODES:
       TasteTester::Commands.runchef
     when :upload
       TasteTester::Commands.upload
+    when :profile
+      TasteTester::Commands.profile
     else
       logger.error("Invalid mode: #{mode}")
       puts parser
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -377,6 +377,12 @@ MODES:
     end
 
     opts.on(
+      '--use-custom-changeset-hook', 'Find a custom changeset. Works on impact.'
+    ) do
+        options[:use_custom_changeset_hook] = true
+    end
+
+    opts.on(
       '--use-custom-impact-hook', 'Use customized impact hook. Works on impact.'
     ) do
       options[:use_custom_impact_hook] = true
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -90,9 +90,9 @@ MODES:
     cookbooks and roles. It also does a fair amount of sanity checking on
     your repo and you may specify --skip-repo-checks to bypass this.
 
-  profile
+  impact
     Examine local repo changes to determine which roles are potentially
-    affected by the changes. Compares the set of modified cookbooks with the
+    impacted by the changes. Compares the set of modified cookbooks with the
     dependency lists of each role and reports any role which depends on a
     modified cookbook. It is recommended to test your changes on at least one
     server of each role, potentially also on multiple platforms.
@@ -421,8 +421,8 @@ MODES:
       TasteTester::Commands.runchef
     when :upload
       TasteTester::Commands.upload
-    when :profile
-      TasteTester::Commands.profile
+    when :impact
+      TasteTester::Commands.impact
     else
       logger.error("Invalid mode: #{mode}")
       puts parser
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -382,18 +382,6 @@ MODES:
     ) do
       options[:skip_post_test_hook] = true
     end
-
-    opts.on(
-      '--use-custom-impact-hook', 'Use customized impact hook. Works on impact.'
-    ) do
-      options[:use_custom_impact_hook] = true
-    end
-
-    opts.on(
-      '--skip-post-impact-hook', 'Skip post-impact hook. Works on impact.'
-    ) do
-      options[:skip_post_impact_hook] = true
-    end
   end
   # rubocop:enable Metrics/BlockLength
 
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -237,10 +237,10 @@ module TasteTester
       impacted_roles = TasteTester::Hooks.impact_find_roles(changes)
       impacted_roles ||= _find_roles(changes)
 
-
       # Do any post processing required on the list of impacted roles, such
       # as looking up hostnames associated with each role.
       final_impact = TasteTester::Hooks.post_impact(impacted_roles)
+      final_impact ||= impacted_roles
 
       # Print the calculated impact. If a print hook is defined that
       # returns true, then the default print function is skipped.
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -210,5 +210,10 @@ module TasteTester
       logger.error(exception.backtrace.join("\n"))
       exit 1
     end
+
+    def self.profile
+      logger.warn('Profile which roles will be affected by current changes')
+      logger.error('Not yet implemented')
+    end
   end
 end
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -212,6 +212,11 @@ module TasteTester
     end
 
     def self.impact
+      if TasteTester::Config.json
+        logger.error('JSON output format is not yet implemented')
+        exit(1)
+      end
+
       # Use the repository specified in config.rb to calculate the changes
       # that may affect Chef. These changes will be further analyzed to
       # determine specific roles which may change due to modifed dependencies.
@@ -227,21 +232,13 @@ module TasteTester
       changes = _find_changeset(repo)
 
       # Use Knife (or custom logic) to check the dependencies of each role
-      # against the list of changes. `impact_roles` will contian the set
+      # against the list of changes. `impacted_roles` will contian the set
       # of roles with direct or indirect (dependency) modifications.
-      if TasteTester::Config.use_custom_impact_hook
-        impact_roles = TasteTester::Hooks.custom_impact(changes)
-      else
-        impact_roles = _find_impact(changes)
-      end
+      impacted_roles = TasteTester::Hooks.impact_find_roles(changes)
 
       # Do any post processing required on the list of impacted roles, such
       # as looking up hostnames associated with each role.
-      unless TasteTester::Config.skip_post_impact_hook
-        final_impact = TasteTester::Hooks.post_impact(impact_roles)
-      end
-
-      final_impact ||= impact_roles
+      final_impact = TasteTester::Hooks.post_impact(impacted_roles)
 
       # Print the calculated impact. If a print hook is defined that
       # returns true, then the default print function is skipped.
@@ -285,119 +282,10 @@ module TasteTester
       return changeset
     end
 
-    def self._find_impact(changeset)
-      if TasteTester::Config.relative_cookbook_dirs.length > 1
-        logger.error('Knife deps does not support multiple cookbook paths.')
-        logger.error('Flatten the repository or use the custom_impact hook.')
-        exit(1)
-      end
-
-      cookbooks = Set.new(changeset.cookbooks)
-      roles = Set.new(changeset.roles)
-      databags = Set.new(changeset.databags)
-
-      if cookbooks.empty? && roles.empty?
-        logger.warn('No cookbooks or roles have been modified.')
-        return []
-      end
-
-      unless cookbooks.empty?
-        logger.info('Modified Cookbooks:')
-        cookbooks.each { |cb| logger.info("\t#{cb}") }
-      end
-      unless roles.empty?
-        logger.info('Modified Roles:')
-        roles.each { |r| logger.info("\t#{r}") }
-      end
-      unless databags.empty?
-        logger.info('Modified Databags:')
-        databags.each { |db| logger.info("\t#{db}") }
-      end
-
-      return _find_impact_roles(cookbooks, roles)
-    end
-
-    def self._find_impact_roles(mod_cookbooks, mod_roles)
-      # Now that we have a list of changed cookbooks and roles, we can check
-      # which roles contain modified dependencies. For simplicity, we will
-      # check using brute force by iterating through each role and comparing
-      # its dependencies with the set of modified files. If a match exists,
-      # the role was impacted, otherwise we can move on.
-      # This may change in the future if it is a perfomance bottleneck.
-
-      role_dir = TasteTester::Config.role_dir
-      chef_dir = File.join(TasteTester::Config.repo,
-                           TasteTester::Config.base_dir)
-
-      # CLI options for knife call
-      config = "--config #{TasteTester::Config.knife_config}"
-      chef_path = "--chef-repo-path #{chef_dir}"
-      options = '--tree --recurse'
-
-      # shell out to knife once and parse the resulting file for dependents
-      # if knife did not exit with 0, print whatever it returned and exit
-      logger.info('Finding dependencies (this may take a minute or two)...')
-      knife = Mixlib::ShellOut.new(
-        "knife deps /#{role_dir}/*.rb #{options} #{config} #{chef_path}",
-      )
-      knife.run_command
-      knife.error!
-
-      # create a hash between roles and their dependencies
-      logger.info('Processing dependencies...')
-      deps_tree = _build_dependency_tree(knife.stdout)
-
-      # compare the dependencies of each role to the list of modified
-      # cookbooks, recording the role as impacted if a match exists
-      impact_roles = Set.new(mod_roles.map(&:name))
-      deps_tree.each do |role, deplist|
-        mod_cookbooks.each do |cb|
-          if deplist.include?(cb.name)
-            impact_roles.add(role)
-            logger.info("\tFound dependency: #{role} --> #{cb.name}")
-            break
-          end
-        end
-      end
-
-      return impact_roles
-    end
-
-    def self._build_dependency_tree(deps)
-      # We want to record all nested lines for each role. The top-level roles
-      # will not be indented, indicating they have no dependents. Thus, when
-      # we see a line without indentation, we can start adding subsequent
-      # lines to a list for that line (ie a hash of role -> [deps]). Since we
-      # don't care about precise dependencies for cookbooks, we can lump all
-      # of the indented layers together into the list for the single role.
-      # This is not the most efficient method as repeated cookbooks will list
-      # their dependencies multiple times for each role they are part of,
-      # but the parsing is significantly easier than building a proper graph.
-
-      tree = {}
-      curr_role = nil
-
-      deps.each_line do |line|
-        elem = line.rstrip
-
-        if elem.length == elem.lstrip.length
-          curr_role = elem
-          tree[curr_role] = Set.new()
-        else
-          tree[curr_role].add(File.basename(elem, File.extname(elem)))
-        end
-      end
-
-      return tree
-    end
-
     def self._print_impact(final_impact)
       if TasteTester::Config.json
-        puts 'JSON output not yet supported'
-      end
-      # TODO: convert to an elsif when JSON is supported,
-      # otherwise fall through to normal output
-      if final_impact.empty?
+        # TODO: parse and print final_impact as JSON for downstream processing
+      elsif final_impact.empty?
         logger.warn('No impacted roles were found.')
       else
         logger.warn('The following roles have modified dependencies.' +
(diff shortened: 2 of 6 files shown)


lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+      options = '--tree --recurse'
+
+      # shell out to knife once and parse the resulting file for dependents
+      # if knife did not exit with 0, print whatever it returned and exit
+      logger.warn('Finding dependencies (this may take a minute or two)...')
+      knife = Mixlib::ShellOut.new(
+        "knife deps /#{role_dir}/*.rb #{options} #{config} #{chef_path}",
+      )
+      knife.run_command
+      if knife.error?
```
jaymzh: This whole if statement is just:

```
knife.error!('knife deps failed, please check output below')
```

lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+
+      # create a hash between roles and their dependencies
+      logger.warn('Processing dependencies...')
+      deps_tree = _build_dependency_tree(knife.stdout)
+
+      # search the hash for roles containing any modified files
+      logger.warn('Searching for impacted roles...')
+
+      impact_roles = []
+      roles.each { |r| impact_roles |= r.name }
```
jaymzh: did you mean `impact_roles = roles.map { |r| r.name }` ?

lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+
+      deps_tree.each do |r, d|
```
jaymzh: please use better iterator variable names, role and ... deplist?

lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+        cookbooks.each do |cb|
+          if d.include?(cb.name)
+            impact_roles |= [r]
```
jaymzh: if you're going for uniquness, why not just use a `set` instead of an array?

bin/taste-tester
```diff
@@ -368,6 +375,24 @@ MODES:
...
     end
+
+    opts.on(
+      '--use-custom-changeset-hook', 'Find a custom changeset. Works on impact.'
+    ) do
+      options[:use_custom_changeset_hook] = true
+    end
+
+    opts.on(
+      '--use-custom-impact-hook', 'Use customized impact hook. Works on impact.'
```
jaymzh: I strongly dislike this, it does not follow the model of the plugins here at all.

Instead, I would model this after the existing plugin infrastructure. I would put the code you want in the common plugin, and let people override it..

that said, what would be *better* would be to find the right hook point sin your common one to make it extensible.

You'll note that taste tester doesnt' provide a way for you to simply write your own upload method, it has a pre_ and post_ upload. For generating the client config, we had `test-remote_client_rb_extra_code`.

I'm happy to help find the right way to factor out the code into reusable parts instead of throwaway parts, but I don't know what specific thigns you want to do internally with your plugin for the phased system so I can't help there... so if you wanna describe that I can probably help.

bin/taste-tester
```diff
@@ -368,6 +375,24 @@ MODES:
     ) do
       options[:skip_post_test_hook] = true
     end
+
+    opts.on(
+      '--use-custom-changeset-hook', 'Find a custom changeset. Works on impact.'
```
jaymzh: This... seems all sorts of scary to me. Let people pass in a start/end, but bypassing the changset calculation doesn't seem like a good idea to me. That's the whole point of TT/BM/GD. What is it you want to do here, we can just extend BM almost certainly...

taste_tester.gemspec
```diff
@@ -14,7 +14,7 @@
 
 Gem::Specification.new do |s|
   s.name = 'taste_tester'
-  s.version = '0.0.13'
+  s.version = '0.0.14'
```
jaymzh: don't do this in this PR... when we're ready to release, then we can do that...

jaymzh:
Oh one other thought I had... TT/BM/GD supports databags. FB Prod doesn't use them but FB CPE does, and so do non-FB users... I think it would make sense for `impact` to print out the databags changing.

And... also, `impact` probably should have a `--json` option for machine-parsability for aftertaste...

bin/taste-tester
joshuamiller01: This isn't about phases; it's about being able to override the entire subcommand (`impact`) with custom internal code where applicable.  What's a good pattern for that?

lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+      if TasteTester::Config.use_custom_impact_hook
+        impact_roles = TasteTester::Hooks.custom_impact(changeset)
+      else
+        impact_roles = _find_impact(changeset)
+      end
+
+      if impact_roles.empty?
+        logger.warn('No impacted roles were found.')
+      else
+        logger.warn('The following roles have modified dependencies. ' +
```
joshuamiller01: You're using `warn` in a lot of places where it's not a warning; this should probably be `info`, or perhaps just using `puts` if this is the "expected output of the command"  as opposed to logging information.

lib/taste_tester/commands.rb
```diff
@@ -210,5 +211,187 @@ def self.upload
...
+        logger.error('Flatten the repository or use the custom_impact hook.')
+        exit(1)
+      end
+
+      cookbooks = Set.new(changeset.cookbooks)
+      roles = Set.new(changeset.roles)
+
+      if cookbooks.empty? && roles.empty?
+        logger.error('No cookbooks or roles have been modified.')
+        exit(1)
```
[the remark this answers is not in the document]
joshuamiller01: Yea, just return here.

lib/taste_tester/commands.rb
jaymzh: In GD, only warn+ was getting sent to output, don't remember if that's the same in TT or not...

bin/taste-tester
jaymzh: @joshuamiller01  why does the internal code need to be different? Without knowing what you want to do internally, it's hard to answer the question. It feels like if the ENTIRE subcommand needs to be different, then (1) the external command isn't that useful OR (2) the external command isn't factored well. The right pattern, were possible, is to have the right hooks throughout the external command so that plugins can shape it's behavior.

joshuamiller01: We can use internal tools to more efficiently determine the affected roles.  Ideally those are open sourced some day, but since they're not in ruby it'll be nontrivial and isn't in scope here.

If the factoring is the problem, them maybe we should have a hook for setting the mechanism which resolves the deps, but the portion which looks at the repository contents and builds the set of affected cookbooks can be standard both internally and open source.   @emroch what do you think?

lib/taste_tester/commands.rb
emroch: TT only outputs warn+ by default.  I was mostly following the example of the rest of the commands and prod_hooks, but I can drop some of these down to info or puts.

lib/taste_tester/commands.rb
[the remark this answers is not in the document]
emroch: Agreed, I will change this to return an empty set.

> Oh one other thought I had... TT/BM/GD supports databags. FB Prod doesn't use them but FB CPE does, and so do non-FB users... I think it would make sense for `impact` to print out the databags changing.
> 
> And... also, `impact` probably should have a `--json` option for machine-parsability for aftertaste...

Good point. I will look at adding databag support and a json option

emroch:
> Oh one other thought I had... TT/BM/GD supports databags. FB Prod doesn't use them but FB CPE does, and so do non-FB users... I think it would make sense for `impact` to print out the databags changing.
> 
> And... also, `impact` probably should have a `--json` option for machine-parsability for aftertaste...

Good point, I will look into adding databag and json support.

lib/taste_tester/commands.rb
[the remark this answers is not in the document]
emroch: Agreed. I will change this to return an empty set.

lib/taste_tester/commands.rb
emroch: `impact_roles = []` is a set, no?  Then the `|=` operator merges the single element set `[r]` with the current `impact_roles` set.  Is there a better way to do this?

lib/taste_tester/commands.rb
emroch: I still want `impact_roles` to be a set. Is there a better way than `impact_roles = Set.new(roles.map { |r| r.name })`?

bin/taste-tester
emroch: The idea here was to allow users to define their own logic for calculating what has been changed.  However, I think you are correct that this is already covered by the Config.repo and a start/end ref.  I have removed the hook and added an additional config option.

lib/taste_tester/commands.rb
emroch: ShellOut#error! does not take any arguments, so I would have to go with the default `Expected process to exit with #{valid_exit_codes.inspect}, but received '#{exitstatus}'`
This is probably ok though since my message wasn't much more helpful.

lib/taste_tester/commands.rb
emroch: Actually, `impact_roles.add(r)` is probably better here

bin/taste-tester
jaymzh: Sure, so having something like TT build a hash object of the source changes and their dependencies, and then hand that too like a `post_process_changes` hook which can than take that and do lookups for like what systems run those cookbooks/roles/whatever.

Then you can have something like a `format_changes` hook, which takes the hash and either returns `nil` if it'd rather the TT format the changes or returns it's own string if it wants to format it itself.

Something like that?

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,179 @@ def self.upload
...
+      )
+      if repo && !repo.exists?
+        fail "Could not open repo from #{TasteTester::Config.repo}"
+      end
+
+      # Compute the set of changes using Between Meals
+      changeset = _find_changeset(repo)
+
+      # Use the changeset computed earlier to find modified roles
+      if TasteTester::Config.use_custom_impact_hook
```
jaymzh: Instead why not just do:

```
TasteTester::Hooks.impact_post_process(_find_impact(changeset))
```

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,179 @@ def self.upload
...
+          :databag_dir =>
+            TasteTester::Config.relative_databag_dir,
+        },
+        @track_symlinks,
+      )
+
+      return changeset
+    end
+
+    def self._find_impact(changeset)
```
jaymzh: is the whole point of _find_impact just to turn these into a set and then print the list of modified roles and cookbooks? And _find_impact_roles is... actually "expand_dependencies" ?

It seems to me a cleaner API would be something like:

```
changes = _find_changeset
full_changes = expand_dependencies(changes) # make this not print things, just build the data
TasteTester::Hooks.post_process_changes!(full_changes) # does nothing in default case, modifies full_changes if desired
# also probably no reason to have a skip option for ^^
unless TasteTester::Hooks.print_changes(full_changes) # returns `true` if it printed anything and `nil` otherwise
  _print_changes(full_changes)
end
```

jaymzh:
(just to be clear that API is based on some guesses about what you're trying to do with this via looking up recipes/roles in scuba and then formatting them for various things, but without knowing specifically what you're impact code shoudl do I don't know that exact API works for you... it's just trying to show the kind of thing I'm thinking of)

lib/taste_tester/commands.rb
emroch: Instead of what? Do you mean in place of the custom_impact hook?

jaymzh: yeah. See the expanded version of this in the other comment.

emroch:
Yeah, so the goal is to report a list of hosts that could be affected by changes to the Chef repo.  The most basic version of this is to report which roles are modified (either directly or through dependencies) and leave it up to the user to determine which hosts are running those roles.  This is where the `post_impact` hook comes in.  Part of that hook could be querying a database of hostnames or doing more specific profiling (i.e. through scuba) to recommend a useful and complete set of hosts to test on.

The actual meat of the impact mode (currently `_find_impact --> _find_impact_roles`) is to transform the set of changes from BM into a set of roles which need to be tested.  The default way to do this is with `knife deps` which builds the recursive dependency tree for all roles and compares against the set of changed cookbooks.  However, knife deps does not work when the cookbooks are nested in subfolders, so we need some way to override this functionality internally.  We also have an internal tool that can calculate the reverse dependencies faster than knife, so we are using that.

So the API should be something like:
```
changes = _find_changeset
# use either the default or custom logic to determine impacted roles
roles_to_test = (use_custom_impact_hook) ? Hooks::custom_impact(changes) : _find_impact(changes)
# do any post processing, such as hostname lookup
post_impact(roles_to_test) unless skip_post_impact_hook
```

What were you going for with the _print_changes function?

jaymzh:
Right, so if i think about how this could be used by various organizations, some people may want a list of roles, others may want a list of "types of hosts" (tier/hostnameprefixe to FB), others may want an actual list of *hosts*, others may want a sample of a list of actual hosts. Depending on that, the output is gonna be VERY different, which is why I made the `_print_changes` hook.

So given what you said I kinda think about it this way:

```
# Simple BM changeset
changes = _find_changeset
# dependency resolution
full_changes = _expand_changset(changes)
```

Now, if you're using `opstool` for dependency resolution, then `_expand_changeset` can use a configurable command. By default it's `knife deps`, but make it an option in the config file. So that FB's config file looks like:

```
dependency_resolution_cmd 'opstool graph -t roles ...'
```

The command can have documented expanders (like `%s` or whatever). Then there doesn't need to be a custom_impact method, it is just nice and configurable.

If that does not work for some reason, then that's fine, `_expand_changeset` would just be a no-op for FB, and you'd call `opstool` inside of `_post_process_impact`.

Then the rest of the code goes:

```
# If the user wants to turn the list of roles/cookbooks into hosts or tiers or wahtever, this is where that 
# goes. No need for a `--skip-post-impact` here - the default hook does nothing
Hooks::post_process_impact!(full_changes)

# If the user changed the structure of full_changes then they'll have their own way of
# outputting it, so call their function first...
ret = Hooks::print_impact(full_changes)

# And if it didn't, then we can go ahead and print it
unless ret
  _print_impact
end
```

emroch:
Ah, I see. The problem with having a configurable command is that the output parsing, not just the printing, is tied to the command.  So if we allow the user to change the command from `knife deps...` to something else, then they likely also need to define their own parsing logic and probably their own printing logic.  At that point, they've just overridden the entire impact function.

bin/taste-tester
emroch: Yes, but the calculating of the dependencies is where the internal and external paths diverge.  Knife should work for the majority of cases, even if it is relatively slow.  Our internal tool is both faster and understands our multilayer repository, which Knife cannot do.  I agree that passing the output of the dependency finder to a post_process hook is the way to go, as then each user can configure the output to be useful (hosts, tiers, roles, etc.).

taste_tester.gemspec
emroch: Reverted. I am testing locally with a different version number though, so please do a sanity check before the final merge that I don't accidentally push this again.

emroch:
@jaymzh How about this:
`impact` would have 3 (really 2.5) places for users to define their own logic.  The first would be resolving the dependencies (more on this in a bit), the second is post-processing, and the third is a custom print function (which is kind of lumped in with the post-processing).  So the overall function would be something like this:
```ruby
repo = BM::Repo.get()

changes = _find_changeset(repo) # common logic, customizable via config options

impacted_roles = impact_resolve_deps(changes) # hook function with default behavior, overridable for custom situations

final_impact = post_impact(impacted_roles) # resolve hostnames, etc. Does nothing by default.

unless Hooks::print_impact(final_impact) # call custom print hook. If not implemented, call default print function
  _print_impact(final_impact)
end
```

The change here is that instead of having a `--use-custom-impact-hook` setting (which really only affects the dependency search logic, so might be better named as `--use-custom-dependency-check`), we would have some hook which has the default behavior of the current dependency checking.  Users who wish/need to do something special (such as use an internal tool), would simply overwrite this hook in their plugin file.  This also has the benefit of eliminating a CLI option which likely won't change between invocations.

So the final hooks file would be (names are flexible if you prefer a particular pattern)
```ruby
def self.impact_resolve_deps(changeset)
  do_stuff
end

def self.impact_post_process(_impact_roles); end

def self.print_impact(_final_impact); end
```

emroch:
> Oh one other thought I had... TT/BM/GD supports databags. FB Prod doesn't use them but FB CPE does, and so do non-FB users... I think it would make sense for `impact` to print out the databags changing.

Can databags be resolved to roles that depend on them?  As far as I can tell, databag items are dynamically loaded by recipes and knife does not include them in dependencies.  Since they are loaded when the recipe is evaluated, theres not really a way to statically analyze where they will be used, and which recipes/roles will be affected by changes.  You could take a pessimistic approach and say that any recipe which mentions a modified databag may be impacted, but that would require searching each recipe's contents.

I have added the databag_dir to the changeset calculation, and I can print out a set of modified databags the same way I report cookbooks and roles (or output via JSON), but I don't see a way to connect the databags to roles.  Please correct me if I am wrong.

lib/taste_tester/commands.rb
```diff
@@ -294,19 +294,24 @@ def self._find_impact(changeset)
...
       end
       unless roles.empty?
-        logger.warn('Modified Roles:')
-        roles.each { |r| logger.warn("\t#{r}") }
+        logger.info('Modified Roles:')
+        roles.each { |r| logger.info("\t#{r}") }
+      end
+      unless databags.empty?
+        logger.info('Modified Databags:')
+        databags.each { |db| logger.info("\t#{db}") }
```
jaymzh: You're 100% correct databags are not gonna help with the dependency tree (they're dynamically called from recipes), however, since you have this code here which prints all "modified" things, it seems useful to support databags (as you've added here)

lib/taste_tester/commands.rb
```diff
@@ -388,7 +393,7 @@ def self._build_dependency_tree(deps)
 
     def self._print_impact(final_impact)
       if TasteTester::Config.json
-        puts "JSON output not yet supported"
+        puts 'JSON output not yet supported'
```
jaymzh: This should return or exit with an error... and in fact in main when you do option parsing, if you're not gonna support JSON for impact, check if those two options were specified together and exit there before you do all this other work.

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,200 @@ def self.upload
...
+      # of roles with direct or indirect (dependency) modifications.
+      if TasteTester::Config.use_custom_impact_hook
+        impact_roles = TasteTester::Hooks.custom_impact(changes)
+      else
+        impact_roles = _find_impact(changes)
+      end
+
+      # Do any post processing required on the list of impacted roles, such
+      # as looking up hostnames associated with each role.
+      unless TasteTester::Config.skip_post_impact_hook
```
jaymzh: I don't think this warrants a flag. The reason some hooks have such a flag is because they're primary validation hooks which someone might need to bypass, but in this case it doesn't make sense and having this combination of "use this" and "don't use this other part even though I forcefully enabled a related part" It hink is just confusing.

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,200 @@ def self.upload
...
+        fail "Could not open repo from #{TasteTester::Config.repo}"
+      end
+
+      changes = _find_changeset(repo)
+
+      # Use Knife (or custom logic) to check the dependencies of each role
+      # against the list of changes. `impact_roles` will contian the set
+      # of roles with direct or indirect (dependency) modifications.
+      if TasteTester::Config.use_custom_impact_hook
+        impact_roles = TasteTester::Hooks.custom_impact(changes)
```
jaymzh: What if we treat this the same we do `print_impact` below.

So we do:

impact_roles = TasteTester::Hooks.custom_impact(changes)
unless impact_roles
  _find_impact(changes)
end
```

Then we don't need these extra flags at all. The default plugin returns nil, so we get the default behavior.

File: lib/taste_tester/commands.rb (hunk 13 above)
Comment (emroch) [reply lib/taste_tester/commands.rb]: That was more of a placeholder for testing. I will replace this with an exit() if I don't get around to it before merging.

File: lib/taste_tester/commands.rb (hunk 14 above)
Comment (emroch) [reply lib/taste_tester/commands.rb]: Agreed. I've removed the flag and `post_impact` will always run and will just be a NOP by default.

File: lib/taste_tester/commands.rb (hunk 15 above)
Comment (emroch) [reply lib/taste_tester/commands.rb]: Yeah, that was my thought.  Have the "custom_impact" hook (renamed of course) be the actual functionality and users can override it using the same hook mechanism as the post_impact/print functions.

File: lib/taste_tester/commands.rb [hunk 16]
```diff
@@ -210,5 +210,88 @@ def self.upload
...
+      # of roles with direct or indirect (dependency) modifications.
+      impacted_roles = TasteTester::Hooks.impact_find_roles(changes)
+
+      # Do any post processing required on the list of impacted roles, such
+      # as looking up hostnames associated with each role.
+      final_impact = TasteTester::Hooks.post_impact(impacted_roles)
+
+      # Print the calculated impact. If a print hook is defined that
+      # returns true, then the default print function is skipped.
+      unless TasteTester::Hooks.print_impact(final_impact)
```
jaymzh: we should be consistent here. Either we do this one like you did `impact_find_roles` above, where the default code is in the base plugin **or** we do `find_impact_roles` the way we did this one, where if the plugin is a no-op we call an internal one...  I *think* I prefer this one to the other one, but I don't feel strongly.

emroch: When you say "this one," do you mean having a no-op that triggers a private, default function?  I think it probably makes sense to have the actual functionality stay in the commands.rb file and have hooks.rb be stubs.

jaymzh: Yes, exactly :)

lib/taste_tester/commands.rb
emroch: This has been replaced with an `exit(1)` at the beginning of `impact`.  I am going to leave it as a TODO/"exercise for the reader" for now unless you feel strongly about it.

emroch:
I am going to leave JSON output as a TODO for later, unless you think it is necessary for this PR.  The infrastructure is there if a user wants to output their own JSON via the `print_impact` hook, and either of the other impact hooks can be used to format the data appropriately.

@jaymzh, @NaomiReeves if everything looks good to you, I think this is ready to go.

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,166 @@ def self.upload
...
+        @track_symlinks,
+      )
+
+      return changeset
+    end
+
+    def self._find_roles(changes)
+      if TasteTester::Config.relative_cookbook_dirs.length > 1
+        logger.error('Knife deps does not support multiple cookbook paths.')
+        logger.error('Flatten the repo or override the resolve_deps function.')
```
jaymzh: that's not the name of this function anymore :)

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,166 @@ def self.upload
...
+      # This creates a recursive tree structure that is then searched for
+      # instances of modified cookbooks. This can be slow since it must read
+      # every line of the Knife output, then search all roles for dependencies.
+      # If you have a custom way to calculate these reverse dependencies, this
+      # is the part you would replace.
+      logger.info('Finding dependencies (this may take a minute or two)...')
+      knife = Mixlib::ShellOut.new(
+        "knife deps /#{TasteTester::Config.role_dir}/*.rb" +
+        " --config #{TasteTester::Config.knife_config}" +
+        " --chef-repo-path #{TasteTester::Config.absolute_base_dir}" +
```
jaymzh: we should file a bug upstream to knife - it's common to have multiple cookbook dirs, and `knife deps` should support this. Can you also do that please?

jaymzh: we should file a bug upstream to knife - it's common to have multiple cookbook dirs, and `knife deps` should support this. Can you also do that please?

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,166 @@ def self.upload
...
+      end
+
+      return impacted_roles
+    end
+
+    def self._print_impact(final_impact)
+      if TasteTester::Config.json
+        # TODO: parse and print final_impact as JSON for downstream processing
+        logger.error('JSON output format is not yet implemented')
+        exit(1)
```
jaymzh: puts JSON.pretty_generate(final_impact)

lib/taste_tester/commands.rb
emroch: Yes, I will submit an issue this afternoon.

lib/taste_tester/commands.rb
emroch: Good catch :)

lib/taste_tester/commands.rb
```diff
@@ -210,5 +210,165 @@ def self.upload
...
+      )
+
+      return changeset
+    end
+
+    def self._find_roles(changes)
+      if TasteTester::Config.relative_cookbook_dirs.length > 1
+        logger.error('Knife deps does not support multiple cookbook paths.')
+        logger.error('Please flatten the cookbooks into a single directory' +
+                     ' or override the impact_find_roles function.')
```
jaymzh: " or define the impact_find_roles method in a local plugin"

emroch: Hopefully that error message wont live long :)
(https://github.com/chef/chef/issues/8651)

PR #113, opened 2019-06-13
Support `bundle` mode with tar transport

malmond77:
Regular taste-tester uses chef-zero directly with knife upload. This has two issues:

1. The chef-client downloads resources one at a time. i.e. each file in a cookbook, each role, etc. This leads to a thundering herd of tiny transactions when taste-testing > 100 hosts in parallel.
2. knife upload is very slow for a large number of cookbook. The cold upload speed for cookbooks is around 1 minute 20s which is significant when given a deadline for testing changes.

This PR adds a new mode where the same chef-zero instance is used, but as a dumb web-server for a single file. This single file is created from the same list of cookbooks, roles and databags as knife would use. 

A single file is much easier to serve: in preliminary benchmarks, the WEBrick backend to chef-zero handles a few, larger downloads much better than a lot of tiny files, especially with TLS.

As this is creating an archive, this only works in a 'full upload' way - there's no distinct advantage to a partial upload. The same set of data that took 1 minute 20s to upload now takes about 1.8s. This is actually a little faster than using between-meals to produce a delta.

The target host's `client.rb` is modified to enable `solo` and `local_mode`, and the `chef_repo_path` is pointed at a temporary `/run/chef/taste-tester` directory. curl + tar are used to download the cookbooks+roles+databags on each run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ you want to test on, i.e. SSH public/private keys, SSH certificates, Kerberos
 * Mixlib::Config
 * Colorize
 * BetweenMeals
+* Minitar
 
 ## Automatic Untesting
 
@@ -110,6 +111,10 @@ The following options are also available:
   `#{ENV['HOME']}/.chef/taste-tester-ref.txt`
 * checksum_dir - The checksum directory to put in knife.conf for users. Default:
   `#{ENV['HOME']}/.chef/checksums`
+* bundle - use a single tar.gz file for transporting cookbooks, roles and
+  databags to clients. Experimental.
+  Default: false
+
 
 ## Plugin
 
--- a/lib/taste_tester/client.rb
+++ b/lib/taste_tester/client.rb
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+require 'minitar'
+require 'find'
 require 'taste_tester/logging'
 require 'between_meals/repo'
 require 'between_meals/knife'
@@ -125,7 +127,88 @@ module TasteTester
 
     private
 
+    def populate(stream, writer, path, destination)
+      full_path = File.join(File.join(TasteTester::Config.repo, path))
+      return unless File.directory?(full_path)
+      # everything is relative to the repo dir. chdir makes handling all the
+      # paths within this simpler
+      Dir.chdir(full_path)
+      Find.find('.') do |p|
+        # ignore current directory. The File.directory? would also skip it,
+        # but we need to do it early because the string is too short for the
+        # next statement.
+        next if p == '.'
+        # paths are enumerated as relative to the input path '.', so we get
+        # './dir/file'. Stripping off the first two characters gives us a
+        # a cleaner 'dir/file' path.
+        name = File.join(destination, p[2..-1])
+        if File.directory?(p)
+          # skip it. This also handles symlinks to directories which aren't
+          # useful either.
+        elsif File.symlink?(p)
+          # tar handling of filenames > 100 characters gets complex. We'd use
+          # split_name from Minitar, but it's a private method. It's
+          # reasonable to assume that all symlink names in the bundle are
+          # less than 100 characters long. Long term, the version of minitar
+          # in chefdk should be upgraded.
+          fail 'Add support for long symlink paths' if name.size > 100
+          # The version of Minitar included in chefdk does not support
+          # symlinks directly. Therefore we use direct writes to the
+          # underlying stream to reproduce the symlinks
+          symlink = {
+            :name => name,
+            :mode => 0644,
+            :typeflag => '2',
+            :size => 0,
+            :linkname => File.readlink(p),
+            :prefix => '',
+          }
+          stream.write(Minitar::PosixHeader.new(symlink))
+        else
+          File.open(p, 'rb') do |r|
+            writer.add_file_simple(
+              name, :mode => 0644, :size => File.size(r)
+            ) do |d, _opts|
+              IO.copy_stream(r, d)
+            end
+          end
+        end
+      end
+    end
+
+    def bundle_upload
+      dest = File.join(@server.bundle_dir, 'tt.tgz')
+      begin
+        Tempfile.create(['tt', '.tgz'], @server.bundle_dir) do |tempfile|
+          stream = Zlib::GzipWriter.new(tempfile)
+          Minitar::Writer.open(stream) do |writer|
+            TasteTester::Config.relative_cookbook_dirs.each do |cb_dir|
+              populate(stream, writer, cb_dir, 'cookbooks')
+            end
+            populate(
+              stream, writer, TasteTester::Config.relative_role_dir, 'roles'
+            )
+            populate(
+              stream, writer, TasteTester::Config.relative_databag_dir,
+              'databag'
+            )
+          end
+          stream.close
+          File.rename(tempfile.path, dest)
+        end
+      rescue Errno::ENOENT
+        # Normally the temporary file is renamed to the dest name. If this
+        # happens, then the cleanup of of the temporary file doesn't work,
+        # but this is fine and expected.
+        nil
+      end
+    end
+
     def full
+      if TasteTester::Config.bundle
+        bundle_upload
+        return
+      end
       logger.warn('Doing full upload')
       @knife.cookbook_upload_all
       @knife.role_upload_all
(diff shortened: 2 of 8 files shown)


lib/taste_tester/client.rb
```diff
@@ -125,14 +125,62 @@ def upload
...
+      # symlinks
+      [out.join(' '), full_paths]
+    end
+
     def full
+      if TasteTester::Config.bundle
+        dest = "#{@server.bundle_dir}/tt.tgz"
+        excludes = []
+        cmd = [
+          [TasteTester::Config.relative_cookbook_dirs, 'cookbooks'],
```
jaymzh: a name.... and it's data.... would be..... a hash.

```
cmd = {
  'cookbooks' => TasteTester::Config.relative_cookbook_dirs,
  'roles' => [TasteTester::Config.relative_role_dir],
  'databags' => [TasteTester::Config.relative_databag_dir],
}.each do |name, list|
```

lib/taste_tester/client.rb
```diff
@@ -125,14 +125,62 @@ def upload
...
+        ].map do |list, name|
+          subcmd, paths = tar_content(list, name)
+          excludes.push(paths)
+          subcmd
+        end.join(' ')
+
+        # use find to locate all symlinks so they can be excluded, and pass
+        # into --exclude-from=- (stdin). Maxdepth of 1 so symlinks used within
+        # cookbooks etc are not excluded.
+        cmd = "find #{excludes.join(' ')} -maxdepth 1 -type l | " +
```
jaymzh: using cmd to make cmd is ugly. It works because ruby is smart, but use different variables.

lib/taste_tester/client.rb
```diff
@@ -125,14 +125,62 @@ def upload
...
 
+    def tar_content(list, name)
+      full_paths = list.map do |path|
+        "#{TasteTester::Config.repo}/#{path}"
+      end
+      out = full_paths.map do |path|
+        if path.start_with?('/')
+          path = path[1..-1]
+        end
+        "--transform 's,#{path},#{name},'"
```
jaymzh: This whole method needs lots of comments... 

The method as a whole does - "tar_content" implies it returns the contents of a tar, which this does not, so what does this do? (yes I know it generates parts of the tar command, sorta, but a comment needs to explain that).

Then a comment at each block... there's a lot of dense stuff here...

lib/taste_tester/client.rb
```diff
@@ -125,14 +125,62 @@ def upload
...
+        end,
+      )
+      # partial tar command line + list of full path used for finding
+      # symlinks
+      [out.join(' '), full_paths]
+    end
+
     def full
+      if TasteTester::Config.bundle
+        dest = "#{@server.bundle_dir}/tt.tgz"
```
jaymzh: I do like the idea of keeping it always having one name, so we keep the same semantics of "an upload changes the state to the current repo"

malmond77:
@jaymzh : Thanks for the feedback. I've substantially changed the methods you wrote comments on. As a proof of concept I used shelling out to CLI tools. I've since replaced everything that is platform dependent with gems available in their environments (Minitar in chefdk, and Gem::Package::TarReader, and Zlib for both).

lib/taste_tester/client.rb
```diff
@@ -125,14 +127,93 @@ def upload
...
+        end
+      end
+    end
+
+    def bundle_upload
+      dest = File.join(@server.bundle_dir, 'tt.tgz')
+      begin
+        Tempfile.create(['tt', 'tgz'], @server.bundle_dir) do |f|
+          stream = Zlib::GzipWriter.new(f)
+          Minitar::Writer.open(stream) do |w|
```
jaymzh: s/w/writer/

lib/taste_tester/client.rb
```diff
@@ -125,14 +127,93 @@ def upload
...
+            TasteTester::Config.relative_cookbook_dirs.each do |c|
```
jaymzh: s/c/cb_dir/

README.md
```diff
@@ -110,6 +111,10 @@ The following options are also available:
   `#{ENV['HOME']}/.chef/taste-tester-ref.txt`
 * checksum_dir - The checksum directory to put in knife.conf for users. Default:
   `#{ENV['HOME']}/.chef/checksums`
+* bundle - use a single tar.gz file for transporting cookbooks, roles and
```
jaymzh: I would add some verbiage here about how this is intended for parallel testing if the same set of sources, but is *not* intended for use for general testing.

lib/taste_tester/client.rb
```diff
@@ -125,14 +127,93 @@ def upload
...
+        next if p == '.'
+        name = File.join(destination, p[2..-1])
+        if File.directory?(p)
+          # skip it. This also handles symlinks to directories which aren't
+          # useful either.
+        elsif File.symlink?(p)
+          # tar handling of filenames > 100 characters gets complex. We'd use
+          # split_name from Minitar, but it's a private method. It's
+          # reasonable to assume that all symlink names in the bundle are
+          # less than 100 characters long. Long term, the version of minitar
```
jaymzh: check if chefdk3 or 4 has a newer one (fb is necessarily on chefdk2)... if not, then file a request upstream.

lib/taste_tester/client.rb
malmond77: This was removed

lib/taste_tester/client.rb
malmond77: This code was removed

lib/taste_tester/client.rb
malmond77: No action needed here

lib/taste_tester/client.rb
```diff
@@ -125,14 +127,93 @@ def upload
...
 
+    def populate(stream, writer, path, destination)
+      full_path = File.join(File.join(TasteTester::Config.repo, path))
+      return unless File.directory?(full_path)
+      # everything is relative to the repo dir. chdir makes handling all the
+      # paths within this simpler
+      Dir.chdir(full_path)
+      Find.find('.') do |p|
+        next if p == '.'
+        name = File.join(destination, p[2..-1])
```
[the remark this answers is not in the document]
malmond77: In next rev

lib/taste_tester/client.rb
malmond77: (y)

lib/taste_tester/client.rb
```diff
@@ -125,14 +127,93 @@ def upload
...
+            end
+          end
+        end
+      end
+    end
+
+    def bundle_upload
+      dest = File.join(@server.bundle_dir, 'tt.tgz')
+      begin
+        Tempfile.create(['tt', 'tgz'], @server.bundle_dir) do |f|
```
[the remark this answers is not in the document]
malmond77: (y)

lib/taste_tester/client.rb
malmond77: (y)

README.md
malmond77: I've added that this is experimental. I think that it is too early to make judgements about whether this is appropriate for general testing. I'll address that in the other comment.

malmond77:
Thanks for all the feedback. On the general comments, I'd like to address each point:

> There's more IO on the destination system
I'm aware that this pre-fetches all the cookbooks to disk, on every run. I had originally chosen `/run/chef` which in all practical circumstances is ramdisk. I decided against it and chose to use `Dir.tmpdir` to ensure this works cross platform. From what I understand: these files are uploaded into a Chef Zero instance running within the main Chef client.

Practically, the amount of I/O involved is trivial - it's about 40MB of files, and a split second of overhead on the run.

> there's more chances you can pull in things you don't intend, or things you intentionally expect to blow up won't.

I tested this fairly extensively. I built a poisoned cookbook with attributes, libraries, resources and recipes that would instantly fail a run if touched. I believe the Chef client strictly obeys role->cookbook references and cookbook->cookbook dependencies in `metadata.rb`. In both this mode, and the regular use of Chef Zero, all cookbooks in the `cookbook_dirs` config are uploaded.

> Further, in FB's specific case, I know Aftertaste doesn't yet support the secure clusters, but this will absolutely not work in that case (well it'll technically work, but it would violate various things you don't want to violate), so be very careful in how you deploy this internally.

Going back to the previous point: taste-tester uploads *all* the cookbooks in `cookbook_dirs`. The issue with secure cluster is noted: however there is no difference in behaviour between the two modes.

lib/taste_tester/client.rb
malmond77: This was superseded.

lib/taste_tester/client.rb
malmond77: I looked closer: the newest minitar doesn't explicitly support symlinks either. I think chasing a dependency on a dependency isn't particularly urgent. I'll leave the comment to explain where this might be improved. It's just not urgent at all.

PR #117, opened 2019-07-22
Use chdir in block mode to restore pwd

malmond77:
Passing a relative repo path like `--repo .` caused the program to lose it's place. Using chdir in a block restores the previous working directory each time so this is safer.

Signed-off-by: Matthew Almond <malmond@fb.com>


jaymzh:
<3

PR #131, opened 2019-11-02
Fix the 'run' command

jaymzh:
It was removed in c21fcd55e15b170e96b02113c0bdf11b06f65eb3, but it's 
trivial to fix, so let's keep it.

This PR has two commits, one reverts the above commit and one is the tiny
fix.

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -66,7 +66,8 @@ TLDR; Most common usage is:
   taste-tester untest -s [host] # Put host back in production
                                 #   (optional - will revert itself after 1 hour)
 
-And you're done! See the above wiki page for more details.
+And you're done!
+Note: There may be site specific testing instructions, see local documentation for details.
 
 MODES:
   test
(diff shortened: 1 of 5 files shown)


jaymzh:
OK, this now excises some of the old silly code and just uses Mixlib::ShellOut's streaming feature... but that depends on m between-meals PR: https://github.com/facebook/between-meals/pull/110

bin/taste-tester
```diff
@@ -48,18 +57,16 @@ TLDR; Most common usage is:
...
-    # Run chef and watch it succeed
+  #{format('%-28s', "  #{cmd}")}  # Run chef and watch it succeed
   <#{verify}>
   taste-tester untest -s [host] # Put host back in production
                                 #   (optional - will revert itself after 1 hour)
 
-And you're done!
-Note: There may be site specific changes, e.g. instead of chef-client you may
-use a wrapper. See local documentation for details.
+And you're done! See the above wiki page for more details.
```
NaomiReeves: let's keep a version of the updated message.

```suggestion
And you're done!
Note: There may be site specific testing instructions, see local documentation for details.
```

PR #133, opened 2020-05-21
More robust SSH tunnel handling

jaymzh:
**Summary**

This is based off of https://github.com/facebook/taste-tester/pull/110 .
It has the same basic idea, but is tested and fixes a variety of bugs
both with that PR and with existing code.

The changes are as follows:
* Use HEREDOCs for the embedded shell so we don't have nested-quotes,
  which is easier to read and less error prone
* Stick some optional logging inside the shell so we know when and why
  tunnels are getting turn down
* The tunnel was never able to be killed by `untest` because we were
  jamming a sudo, but the whole thing is run in sudo so we were getting
  `sudo bash -c sudo kill -9 -- -<pid>` which wasn't working. Removing
  the redundant sudo fixed that
* We weren't going high enough the PID-chain to get SSH. Now we are:
    ```
    [phil@scale-lists1 ~]$ cat /etc/chef/test_timestamp
    14144
    [phil@scale-lists1 ~]$ ps -ef | grep 14144
    root     14144  1322  0 02:30 ?        00:00:00 sshd: phil [priv]
    phil     14152 14144  0 02:30 ?        00:00:00 sshd: phil@notty
    phil     15543 31390  0 02:34 pts/1    00:00:00 grep --color=auto 14144
    ````
* The license was typo'd and gem was bitching about it being an invalid
    license
* rugged doens't build on ruby 2.3, so dropped it

It's got some nice logging now that you'll see in the test plan

**Test Plan**

* When I remove the timestamp:
    ```
    May 21 01:54:47 scale-lists1 taste-tester: Ending tunnel: timestamp file disappeared
    ```
* When I change the PID in the timestamp:
    ```
    May 21 01:56:02 scale-lists1 taste-tester: Ending tunnel: timestamp PGID changed
    ```
* When I taste-untest the sshd process goes away

Tested with both root and non-root users, extensively

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,9 @@ language: ruby
 sudo: false
 
 rvm:
-  - 2.3.4
   - 2.4.0
+  - 2.5.3
+  - 2.6.1
 
 gemfile:
   - Gemfile
--- a/circle.yml
+++ b/circle.yml
@@ -16,14 +16,15 @@ jobs:
   ruby_2_5:
     <<: *defaults
     docker:
-      - image: circleci/ruby:2.4
+      - image: circleci/ruby:2.5
   ruby_2_6:
     <<: *defaults
     docker:
-      - image: circleci/ruby:2.4
+      - image: circleci/ruby:2.6
 workflows:
   version: 2
   test:
     jobs:
-      - ruby_2_3
       - ruby_2_4
+      - ruby_2_5
+      - ruby_2_6
--- a/lib/taste_tester/server.rb
+++ b/lib/taste_tester/server.rb
@@ -16,7 +16,6 @@
 
 require 'fileutils'
 require 'socket'
-require 'timeout'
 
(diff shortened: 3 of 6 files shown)


jaymzh:
looks like rugged doesn't work with 2.3 anymore. I'll make a PR to drop 2.3 and add 2.5 and 2.6

jaymzh:
https://github.com/facebook/taste-tester/pull/134 will fix that

jaymzh:
I tried to fix the tests in that other PR, but did it poorly. But whatever, they're fixed in this one now. :)

jaymzh:
this breaks for non-sudo use-case, don't merge. Will push updates when I have all cases solidly tested.

jaymzh:
OK updated with full support for root and non-root cases.

jaymzh:
This should make @mikedodge04 and @michel-slm happy.

PR #138, opened 2020-06-15
Follow symlinked directories in cookbooks and other dirs

malmond77:
We have several distinct trees of cookbooks, some of which share content
via symlinks. `Find.find` in Ruby does not follow symlinks.

Wrap the loop with a queue of directories to look at, starting with none
(i.e. current directory). If we find a symlinked directory, add it as a
prefix to `**/*` which still returns paths relative to the `Dir.chdir`

Use of `Dir.glob` eliminates `p` vs `relative_name` distinctions.

### DIFF
--- a/lib/taste_tester/client.rb
+++ b/lib/taste_tester/client.rb
@@ -136,53 +136,48 @@ module TasteTester
       # everything is relative to the repo dir. chdir makes handling all the
       # paths within this simpler
       Dir.chdir(full_path) do
-        Find.find('.') do |p|
-          # ignore current directory. The File.directory? would also skip it,
-          # but we need to do it early because the string is too short for the
-          # next statement.
-          next if p == '.'
-          # paths are enumerated as relative to the input path '.', so we get
-          # './dir/file'. Stripping off the first two characters gives us a
-          # a cleaner 'dir/file' path.
-          relative_name = p[2..-1]
-          # some exclusions are rooted relative to cookbooks which are one
-          # level below the directory included here. Strip off the first part
-          # of the path. e.g. `fb_cookbook/spec/foo.rb` would be `spec/foo.rb`
-          # which matches `spec/*`.
-          relative_name_minus_first = relative_name.split(
-            File::SEPARATOR,
-          )[1..-1].join(File::SEPARATOR)
-          next if chefignores.ignored?(relative_name) ||
-            chefignores.ignored?(relative_name_minus_first)
-          name = File.join(destination, relative_name)
-          if File.directory?(p)
-            # skip it. This also handles symlinks to directories which aren't
-            # useful either.
-          elsif File.symlink?(p)
-            # tar handling of filenames > 100 characters gets complex. We'd use
-            # split_name from Minitar, but it's a private method. It's
-            # reasonable to assume that all symlink names in the bundle are
-            # less than 100 characters long. Long term, the version of minitar
-            # in chefdk should be upgraded.
-            fail 'Add support for long symlink paths' if name.size > 100
-            # The version of Minitar included in chefdk does not support
-            # symlinks directly. Therefore we use direct writes to the
-            # underlying stream to reproduce the symlinks
-            symlink = {
-              :name => name,
-              :mode => 0644,
-              :typeflag => '2',
-              :size => 0,
-              :linkname => File.readlink(p),
-              :prefix => '',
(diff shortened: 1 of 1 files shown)


jaymzh:
Just to be clear, symlinks inside of cookbooks should *never* be followed - or created. This was teh cause of... I can't even count how many SEV1s and SEV2s at Facebook.

Symlinks should only be followed for cookbooks themselves (or dirs that contain cookbooks)

malmond77:
I looked for examples of this before, didn't find any. I agree so I've made it only work on the top level items (i.e. where `minus_first` is an empty string).

lib/taste_tester/client.rb
```diff
@@ -136,53 +136,48 @@ def populate(stream, writer, path, destination)
...
+              # cookbooks between codebases.
+              if minus_first == '' && File.symlink?(p)
+                look_at.push("#{p}#{File::SEPARATOR}")
+              end
+            elsif File.symlink?(p)
+              # tar handling of filenames > 100 characters gets complex. We'd
+              # use split_name from Minitar, but it's a private method. It's
+              # reasonable to assume that all symlink names in the bundle are
+              # less than 100 characters long. Long term, the version of minitar
+              # in chefdk should be upgraded.
```
davide125, line 160: Can you file an issue upstream for this and reference it in the comment?

jaymzh: Well chefdk is deprecated. I imagine there's vastly newer minitar in Chef Workstation. @NaomiReeves can probably tell you if it's worth filing a bug or if upgrades-coming-soon-to-a-devserver-near-you will cover it.

malmond77: It's been a while since I looked at this - surely the better path is to explicitly depend on the minimum version of gemspec that fixes this. We can have multiple versions of the gem installed, right?

lib/taste_tester/client.rb
```diff
@@ -136,53 +136,48 @@ def populate(stream, writer, path, destination)
...
+            next if chefignores.ignored?(p) ||
+              chefignores.ignored?(minus_first)
+            name = File.join(destination, p)
+            if File.directory?(p)
+              # we don't store directories in the tar, but we do want to follow
+              # top level symlinked directories as they are used to share
+              # cookbooks between codebases.
+              if minus_first == '' && File.symlink?(p)
+                look_at.push("#{p}#{File::SEPARATOR}")
+              end
```
jaymzh, line 154: OK so the old code *already* supported cookbooks themselves being symlinks.

So what I *THINK* you're doing here is stuff like allowing `core` to be a symlink?

lib/taste_tester/client.rb
```diff
@@ -136,53 +136,48 @@ def populate(stream, writer, path, destination)
...
-            stream.write(Minitar::PosixHeader.new(symlink))
-          else
-            File.open(p, 'rb') do |r|
-              writer.add_file_simple(
-                name, :mode => 0644, :size => File.size(r)
-              ) do |d, _opts|
-                IO.copy_stream(r, d)
+        look_at = ['']
+        while (prefix = look_at.pop)
+          Dir.glob("#{prefix}**/*", File::FNM_DOTMATCH) do |p|
```
jaymzh, line 141: 1. Why do you want to match hidden files? Cookbooks can't be named with a leading `.`
2. You're dutifully using `File::SEPARATOR` everywhere... except you have a hard-coded slash here. Wat

lib/taste_tester/client.rb
malmond77: `populate() is called on each of the entries in `cookbook_dirs`, e.g. `path = 'core'`. This is taking top level directories within `core` and resolving symlinked directories.

lib/taste_tester/client.rb
malmond77: `Find.find` matches hidden files, so this is keeping parity. If something shouldn't be uploaded, it should be in chefignore. I think we make use of this e.g. `fb_skel/files/default/skel-e/.bashrc`.

I'll fix the `File::SEPARATOR`.

lib/taste_tester/client.rb
malmond77: This is an issue in this program first and foremost, so I've captured it here: #139. Can we agree to fix this as a separate issue and get this fix in?

PR #140, opened 2020-06-24
Setup github workflow for CI

davide125:
Copypasta from https://github.com/facebook/between-meals/pull/115


.github/workflows/ci.yml
```diff
@@ -0,0 +1,46 @@
+name: Continuous Integration
+on:
+  push:
+    branches: [master]
+  pull_request:
```
jaymzh, line 5: again I'd recommend pullreqeests against master specifically

PR #146, opened 2020-08-24
Fix bug in taste-untester

jaymzh:
* remove unused two variables
* source the config file where it can actually be used


jaymzh:
The lint error here is pre-existing... and I actually fix it in my other PR...

PR #157, opened 2020-12-19
Allow testing and untesting of hosts to be specified as extra arguments

Babar:
'cause I was tired of having to do:
`taste-tester test -ys $( list-webservers | shuf -n 5 | paste -sd, )`
Now I can just do:
`taste-tester test -ys $( list-webservers | shuf -n 5 )`
Most useful when the list comes from a command. No longer need to `| paste -sd,`

One might think it's small, but:
1. Not everyone knows about the `| paste -sd,` trick
2. Storing the hosts in a variable allows easy processing with other tools like `parallel`

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -56,7 +56,7 @@ Usage: taste-tester <mode> [<options>] [optional servers to test/untest]
 
 TLDR; Most common usage is:
   vi cookbooks/...              # Make your changes and commit locally
-  taste-tester test -s [host]   # Put host in test mode
+  taste-tester test [host]      # Put host in test mode
   ssh root@[host]               # Log on host
   #{the_command_shorten_to_28}  # Run chef and watch it break
   vi cookbooks/...              # Fix your cookbooks
@@ -64,7 +64,7 @@ TLDR; Most common usage is:
   ssh root@[host]               # Log on host
   #{the_command_shorten_to_28}  # Run chef and watch it succeed
   <#{verify}>
-  taste-tester untest -s [host] # Put host back in production
+  taste-tester untest [host]    # Put host back in production
                                 #   (optional - will revert itself after 1 hour)
 
 And you're done!
@@ -73,9 +73,9 @@ Note: There may be site specific testing instructions, see local documentation f
 MODES:
   test
     Sync your local repo to your virtual Chef server (same as 'upload'), and
-    point some production server(s) specified by -s (or as arguments) to your
-    virtual chef server for testing.  If you have you a plugin that uses the
-    hookpoint, it'll may amend your commit message to denote the server you tested.
+    point some production server(s) specified as arguments to your virtual
+    chef server for testing.  If you have you a plugin that uses the hookpoint,
+    it'll may amend your commit message to denote the server you tested.
     Returns:
       0 Success.
       1 Failure, e.g. host not reachable or usage error.
@@ -100,17 +100,17 @@ MODES:
     server of each role, potentially also on multiple platforms.
 
   keeptesting
-    Extend the testing time on server specified by -s by 1 hour unless
+    Extend the testing time on the servers passed as arguments by 1 hour unless
     otherwise specified by -t.
 
   untest
-    Return the server(s) specified (in -s or as arguments) to production.
+    Return the server(s) specified as arguments to production.
 
   status
     Print out the state of the world.
 
   run
-    Run #{cmd} on the machine specified by '-s' over SSH and print the output.
+    Run #{cmd} on the machine(s) specified as arguments over SSH and print the output.
     NOTE!! This is #{'NOT'.red} a sufficient test, you must log onto the remote
     machine and verify the changes you are trying to make are actually present.
 
@@ -323,16 +323,8 @@ MODES:
     end
 
     opts.on(
-      '-s', '--split', 'Split server list on commas.'
-    ) do
-      options[:split] = true
-    end
-
-    opts.on(
-      '--servers', 'DEPRECATED: Alias for --split for backward compatibility.'
-    ) do
-      options[:split] = true
-    end
+      '-s', '--servers', 'DEPRECATED: ignored for backward compatibility.'
+    )
 
     opts.on(
       '--user USER', 'Custom username for SSH, defaults to "root".' +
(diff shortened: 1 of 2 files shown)


bin/taste-tester
```diff
@@ -434,6 +435,10 @@ MODES:
     TasteTester::Hooks.get(path)
   end
 
+  if ARGV && [:test, :untest].include?(mode.to_sym)
+    TasteTester::Config.servers.concat ARGV
```
jaymzh, line 439: So now some of your servers are in `TasteTester::Config` and some are in `options[:server]`? Doesn't seem useful. Am I missing the rest of the code where you actually implement looking at `TasteTester::Config.servers`?

In general I dislike this approach. It's brittle.

If we wanted to do something like this, I'd instead suggest accepting `-` as an arg to `-s`, and if that's there reading hostnames from STDIN.

jaymzh (changes requested):

bin/taste-tester
Babar: Those are merged 10 lines above (424 or 425). If you prefer I can move this before the merge and `concat` to `options[:servers]`. Don't really see how that makes any difference, but I guess that means I can remove the default in `TasteTester::Config`
And I know you dislike that approach, you've been refusing to do this for years even though many users have asked for it. I've been using this code for a while and others mentioned that it could be useful for them.

jaymzh: There is no changes on lines 424 or 425. Are you missing changes in your PR? The only other changes in this file in this PR are in the help.

jaymzh: Oh. I see.

I'll let others weigh in. I'm open to the idea of easier ways of passing more servers in, but the idea you can mix-n-match like this is just confusing to users. It's non-standard... I can't think of another utility that works like this. Where as `-s -` is a pretty common pattern.

I'd like it to continue to look like a unix utility, not a babar-specific utility.

malmond77:
I like the idea, but I don't like the ambiguity of the combination. If you specify `-s` or `--split` then take a comma separated list. If you want to use the remaining positional parameters that is great too, but don't allow them to be combined. Make that an error condition. One day we can sunset `-s` too.

davide125:
Looks like we have a few options:
- leave it as is and add `-s -`
- support multiple `-s` args (`-s server1 -s server2,server3`)
- drop `-s` and make the servers list a varargs positional argument thing

These aren't mutually exclusive.

bin/taste-tester
```diff
@@ -418,6 +424,20 @@ MODES:
...
+      if options[:split]
+        options[:servers] = ARGV.map { |s| s.split(/,/) }.flatten
+      else
+        options[:servers] = ARGV
+      end
+    end
+  else
+    puts "Mode #{mode} does not accept extra arguments: #{ARGV}\n\n"
+    puts parser
+    exit(1)
```
jaymzh, line 438: you exist if there's NOT an ARGV? seems wrong.

bin/taste-tester
```diff
@@ -322,10 +323,15 @@ MODES:
...
 
     opts.on(
-      '-s', '--servers SERVER[,SERVER]', Array,
-      'Server to test/untest/keeptesting.'
-    ) do |s|
-      options[:servers] = s
+      '-s', '--split', 'Split server list on commas.'
+    ) do
+      options[:split] = true
+    end
```
jaymzh, line 329: There is no need for such an option. It should certainly just handle the case in which the first argv is coma-separated, or space separated, or some combination thereof. Neither character is allowed in hostnames, so the intended behavior is very clear.

What i don't want is to be passing in sometimes with -s arg and sometimes with positional arguments.

There are two acceptable options here:

1. Accept `-` to `--servers` (preferred) - and actually is better for you since what you're doing is piping in from other commands.
2. Migrate to positional arguments, we can keep `-s` around as a no-op for backcompat.

In both cases `split(/[\s,]+/)` so people can pass in from whatever the fuck they want.

Again, making an option for stdin seems *significantly* better all around.

The only thing that shouldn't happen is different parsing for different ways of passing it in. That's the worst UI possible.

jaymzh (changes requested):

bin/taste-tester
Babar: Oh right that else goes one level up, sorry. Nice catch!

bin/taste-tester
Babar: I think positional arguments are a lot more common. I can even implement a `-` as argument which would read stdin.
But positional arguments feel much logical to me. It's like` grep pattern file1 file2` Or more `systemctl status something somethingelse`
The main problem with stdin is how do you deal with user input?
Like, `are you sure you want to do this?`, or `Do you want to amend your commit message?`
You can't really read the list of hosts from stdin and then read user input
You could read from a file and use `<( ... )` but that is cumbersome

I personally vote for deprecating `-s` entirely, meaning keeping it as a noop for backward compatibility

Babar:
Just noticed @davide125 and @malmond77 comments, sorry!
So yeah, I think we all agree that are several options, some being worse than others.
My personal favorite is the last one @davide125 mentioned (as I wrote in the comment above):
* Drop `-s` and just use positional arguments. Always split them all with `,`, and keep `-s` for a while with an option that does nothing. The current implementation was going half way there, trying to meet both sides, but ended up being wrong.

malmond77:
I'm not in favor of dropping `-s` in one step: this will hard break every single automation (e.g. aftertaste) with no transition possible. I would like to see both supported, in a mutually exclusive way for <some time> so everyone has a chance to update their code. I'm curious - how do we get the message out about this breaking change? Do we set a date where the compatibility is dropped. Can we do better than that somehow?

Babar:
Nobody suggested dropping it in one step. As I wrote:
> keep -s for a while with an option that does nothing

which is basically what the latest commit does: c428a2a
Instead of doing nothing, it could emit a warning, but I worry this might still break some automation.

We could start it like this, announce it somewhere, then in a few months, change that no-op into a deprecation warning, and then a few months later, drop it altogether.

jaymzh:
I mean @malmond77 making it a no-op keeps backwards compat. because `-s foo` in that world sees `foo` as the first positional argument, so it's fine.

OK, I like this code. It's backwards compatible in nice ways, and adds some functionality.

jaymzh (approved):
Would like a 2nd approval before I merge, but I'm good with this.

Thanks @Babar for working through the review.

NaomiReeves (approved):
Make sure that you update the internal wrappers to account for this change, many of them are looking for the `-s` with args. 

I'll build this and roll it out in over the next few weeks.

PR #158, opened 2020-12-19
Allow parallel testing

Babar:
If you have a bunch of hosts being tested, waiting for them sequentially can take a long time, for nothing.
Just use threads to do them all in parallel.

Hopefully this keeps the return codes identical, if anybody relies on them

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -419,7 +419,6 @@ MODES:
       end
       options[:parallel_hosts] = n.to_i
     end
-
   end
 
   if mode == 'help'
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -409,6 +409,17 @@ MODES:
     ) do
       options[:skip_post_test_hook] = true
     end
+
+    opts.on(
+      '--parallel-host-requests NUMHOSTS', 'Start NUMHOSTS threads in parallel'
+    ) do |n|
+      unless n.to_i > 0
+        logger.error("Sorry, number of threads cannot be less than 1: #{n}")
+        exit(1)
+      end
+      options[:parallel_hosts] = n.to_i
+    end
+
   end
 
   if mode == 'help'
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -106,13 +106,13 @@ module TasteTester
         TasteTester::Hooks.post_test(TasteTester::Config.dryrun, repo,
                                      hosts.select { |_, r| r == :ok }.keys)
       end
-      if hosts.values.all? :ok
+      if hosts.values.all?(:ok)
         # No exceptions, complete success: every host listed is now configured
         # to use our chef-zero instance.
         exit(0)
       end
-      if hosts.values.none? :ok
-        if hosts.values.any? :ssh_error
+      if hosts.values.none?(:ok)
+        if hosts.values.any?(:ssh_error)
           # All the hosts we had failed, with at least one because of ssh
           exit(1)
         end
@@ -143,6 +143,7 @@ module TasteTester
           loop do
             begin
               hostname = queue.pop(true)
+              next unless hostname
               Thread.current[:hostname] = hostname
               TasteTester::Host.new(hostname, server).send method
             rescue ThreadError
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -97,13 +97,14 @@ module TasteTester
       end
       unless TasteTester::Config.skip_pre_test_hook ||
           TasteTester::Config.linkonly
-        TasteTester::Hooks.pre_test(TasteTester::Config.dryrun, repo, hosts)
+        TasteTester::Hooks.pre_test(TasteTester::Config.dryrun, repo,
+                                    TasteTester::Config.servers)
       end
       hosts = _run_on_hosts(:test)
       unless TasteTester::Config.skip_post_test_hook ||
           TasteTester::Config.linkonly
         TasteTester::Hooks.post_test(TasteTester::Config.dryrun, repo,
-                                     hosts.select { |h, r| r === :ok }.keys)
+                                     hosts.select { |_, r| r == :ok }.keys)
       end
       if hosts.values.all? :ok
         # No exceptions, complete success: every host listed is now configured
@@ -150,7 +151,6 @@ module TasteTester
           end
         end
       end
-      connect_failures = 0
       host_threads.each do |host_thread|
         result = :unknown_error
         begin
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -142,8 +142,8 @@ module TasteTester
           Thread.current.report_on_exception = false
           loop do
             begin
+              # non_block: true makes pop() throw a ThreadError when empty
               hostname = queue.pop(true)
-              next unless hostname
               Thread.current[:hostname] = hostname
               TasteTester::Host.new(hostname, server).send method
             rescue ThreadError
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -64,8 +64,7 @@ module TasteTester
     end
 
     def self.test
-      hosts = TasteTester::Config.servers
-      unless hosts
+      unless TasteTester::Config.servers
         logger.warn('You must provide a hostname')
         exit(1)
       end
@@ -82,7 +81,6 @@ module TasteTester
         end
         upload
       end
-      server = TasteTester::Server.new
       unless TasteTester::Config.linkonly
         if TasteTester::Config.no_repo
           repo = nil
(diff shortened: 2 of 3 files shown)


lib/taste_tester/commands.rb
```diff
@@ -102,29 +102,40 @@ def self.test
...
+      if tested_hosts.to_set == hosts.to_set
         # No exceptions, complete success: every host listed is now configured
         # to use our chef-zero instance.
         exit(0)
       end
       if tested_hosts.empty?
+        if connect_failures > 0
+          # All the hosts we had failed
+          exit(1)
+        end
```
jaymzh, line 138: hmm not quite true, right? We could have had 1 connection failure and 10 failures due to already-being tested..

jaymzh (changes requested):
This is pretty good and in general I really like this idea.

However, having been an author of https://github.com/txcketmaster-xx/onall/ I know that this can go pretty crazy pretty quickly and kill your local host.

So we should implement a max-parallel set to like 10 or so, which we can adjust with `--parallel-host-requests` or some such

lib/taste_tester/commands.rb
Babar: Yeah, I wanted to try and keep the previous behavior as much as I could.

lib/taste_tester/commands.rb
```diff
@@ -102,29 +102,47 @@ def self.test
...
-      # building tested_hosts from hosts directly.
-      if tested_hosts == hosts
+      if tested_hosts.to_set == hosts.to_set
         # No exceptions, complete success: every host listed is now configured
         # to use our chef-zero instance.
         exit(0)
       end
       if tested_hosts.empty?
+        if connect_failures > 0
+          # All the hosts we had failed, with at least one because of ssh
```
jaymzh, line 143: yeah but we will lose here what @malmond77 / @joshuamiller01  want - whether he needs to just try new hosts because they're already in testing, or if there was some other systemic issue. This is important for `aftertaste`.

With this new code if 9 hosts are already-tested and one has a read-only filesystem or something, instead of aftertaste just picking new hosts (which it should), it'll think they all failed for systemic issues and fail altogether.

So... this will have very adverse affects on aftertaste, so the three of you should sort that out amongst yourselves.

lib/taste_tester/commands.rb
```diff
@@ -102,29 +102,47 @@ def self.test
...
       tested_hosts = []
+      host_threads = []
+      connect_failures = 0
       hosts.each do |hostname|
-        host = TasteTester::Host.new(hostname, server)
+        # Poor man thread pool manager: keeping it simple
+        nb_threads_over_max = host_threads.length - TasteTester::Config.parallel_hosts
+        if nb_threads_over_max >= 0
+          host_threads[nb_threads_over_max].join
+        end
```
jaymzh, line 112: Wait, so if we have 10 max threads, and we've made an 11th, we wait for the 11th to finish? That's... I mean that'll solve for too-many-threads for sure. But it also means that if you have 20 hosts to test and a max conns of 10, you'll do 10 in parallel, then you'll wait for the 11th (which means the other 10 are likely to finish up), and then proceed to do the next 10 one at a time (since the number of elements in the array is still over 10). You'll fire off 12, then wait for 12, then fire off 13 then wait for 13 and so on. That doesn't seem like it does what you want.

Babar: No. It's not pure thread pool, 'cause I didn't want to use a Queue (I could, that would make @rb2k happy).
Basically, if you have 10 threads and you're trying to put the 11th, you'll wait for the **first** one to finish. Then when it finishes, the 12th will wait for the second one, etc...
Yes, in practice, it could very well be that you're waiting on the one thread that's taking the longest, and you could run more. The added complexity did not feel worth it, but I don't mind creating a Queue and having N threads polling from it instead.

lib/taste_tester/commands.rb
Babar: I checked the code, and as of today, `aftertaste` uses Python's asyncio to do this. So basically it runs one `taste-tester` command per host. So this will work, as for one host, you'll get the same exit code.
And honestly, it cannot really change anything, 'cause before it would barf at the first host that failed, so you'd have to check the output or the hosts to figure out which one. Here, if you have one bad hosts at the 5th element of 10, instead of putting the first 4 to testing and crashing returning 1 (default on exception iirc), it will put all 9, and return 2.

lib/taste_tester/commands.rb
Babar: OK, I caved and implemented a thread pool like @rb2k suggested. Still works pretty much the same in my testing :)

Babar:
@jaymzh care to review this again please? Been ready for a while, it now has a proper thread pool using a queue, with a max that's configurable on the command line. And it still works :)

lib/taste_tester/commands.rb
```diff
@@ -99,80 +97,88 @@ def self.test
...
+      hosts_result = {}
+      host_threads = []
+      queue = Queue.new
+      hosts.each { |hostname| queue << hostname }
+      [TasteTester::Config.parallel_hosts, hosts.length].min.times do
+        host_threads << Thread.new do
+          Thread.current.report_on_exception = false
+          loop do
+            begin
+              hostname = queue.pop(true)
```
jaymzh, line 145: do we need like `break unless hostname` here?

jaymzh (commented):
oh sorry, missed this. One comment, but this looks really nice.

bin/taste-tester
```diff
@@ -409,6 +409,16 @@ MODES:
...
     end
+
+    opts.on(
+      '--parallel-host-requests NUMHOSTS', 'Start NUMHOSTS threads in parallel'
+    ) do |n|
+      unless n.to_i > 0
+        logger.error("Sorry, number of threads cannot be less than 1: #{n}")
+        exit(1)
+      end
+      options[:parallel_hosts] = n.to_i
```
jaymzh, line 420: oh onemore comment, please default this to like... 3 or 5 or something small.

lib/taste_tester/commands.rb
```diff
@@ -99,80 +97,88 @@ def self.test
...
           TasteTester::Config.linkonly
         TasteTester::Hooks.post_test(TasteTester::Config.dryrun, repo,
-                                     tested_hosts)
+                                     hosts.select { |_, r| r == :ok }.keys)
       end
-      # Strictly: hosts and tested_hosts should be sets to eliminate variance in
-      # order or duplicates. The exact comparison works here because we're
-      # building tested_hosts from hosts directly.
-      if tested_hosts == hosts
+      if hosts.values.all? :ok
```
jaymzh, line 109: arg methods should have parens `all?(:ok)`

lib/taste_tester/commands.rb
```diff
@@ -99,80 +97,88 @@ def self.test
...
         # No exceptions, complete success: every host listed is now configured
         # to use our chef-zero instance.
         exit(0)
       end
-      if tested_hosts.empty?
+      if hosts.values.none? :ok
```
jaymzh, line 114: same

lib/taste_tester/commands.rb
```diff
@@ -99,80 +97,88 @@ def self.test
...
+        if hosts.values.any? :ssh_error
```
jaymzh, line 115: same

jaymzh (commented):
and a few more style nits

lib/taste_tester/commands.rb
Babar: We fill the queue with hostnames 3 lines above, and if it's empty it will throw and we rescue it.
I don't see how this might be empty, but on the other hand, it doesn't really hurt either :)

bin/taste-tester
Babar: I think I defaulted it to the number of cores, like what parallel does. I can put that lower.

lib/taste_tester/config.rb
```diff
@@ -64,6 +65,7 @@ module Config
     json false
     jumps nil
     windows_target false
+    parallel_hosts Etc.nprocessors
```
Babar, line 68: This is where I default it. The number of cores should be safe for most systems, but happy to put something smaller.

lib/taste_tester/commands.rb
jaymzh: what happens when the queue is empty though? And you're out of work?

lib/taste_tester/commands.rb
```diff
@@ -143,6 +143,7 @@ def self._run_on_hosts(method)
           loop do
             begin
               hostname = queue.pop(true)
+              next unless hostname
```
jaymzh, line 146: why next?? end the thread. `break` the loop, exit the thread. There's no more hostnames in the queue.

Babar: If there is no more hostname in the queue, `pop` throws. This means we un-queued something that was empty:
https://ruby-doc.org/core-2.5.0/Queue.html#method-i-pop
> If the queue is empty, the calling thread is suspended until data is pushed onto the queue. If non_block is true, the thread isn't suspended, and ThreadError is raised.

jaymzh: ah, OK, that's what I was missing. In that case no need to check if hostname is nil.. but put a comment that you're passing `true` to `pop` so that ThreadError will be thrown when the queue is empty. Sorry I didn't understand.

jaymzh (approved):

PR #159, opened 2021-01-19
Re-add integration testing to CI workflow

NaomiReeves:
Somehow this was dropped so I'm re-adding it to the workflow and updating some of the other action steps in our CI workflow.

The upstream source for the action step to run the [rubocop linter as it's own step has been archived](https://github.com/andrewmcodes/rubocop-linter-action), so I'm cleaning that up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,8 @@ jobs:
       run: bundle exec rspec
     - name: Run rubocop
       run: bundle exec rubocop --display-cop-names
+    - name: Run integration tests
+      run: ./test/run.sh
 # This doesn't work on PRs from forks due to GITHUB_TOKEN, which makes
 # it kinda useless for now
 # https://github.com/andrewmcodes/rubocop-linter-action/issues/68
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,18 +25,6 @@ jobs:
       run: bundle exec rubocop --display-cop-names
     - name: Run integration tests
       run: ./test/run.sh
-# This doesn't work on PRs from forks due to GITHUB_TOKEN, which makes
-# it kinda useless for now
-# https://github.com/andrewmcodes/rubocop-linter-action/issues/68
-#  rubocop:
-#    runs-on: ubuntu-latest
-#    steps:
-#    - name: Checkout repository
-#      uses: actions/checkout@v2
-#    - name: Run Rubocop Linter
-#      uses: andrewmcodes/rubocop-linter-action@v3.2.0
-#      env:
-#        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   markdown:
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


jaymzh (approved):

.github/workflows/ci.yml
```diff
@@ -23,24 +23,14 @@ jobs:
       run: bundle exec rspec
     - name: Run rubocop
       run: bundle exec rubocop --display-cop-names
-# This doesn't work on PRs from forks due to GITHUB_TOKEN, which makes
```
jaymzh, line 26: I'd leave this... I think they've fixed this now and I don't wanna re-remember how to do this :)

NaomiReeves: This wasn't fixed on the upstream repo and as I note in my summary, [it was ultimately archived as a result of it]( https://github.com/andrewmcodes/rubocop-linter-action). Or do you want to just see the attempt to do this?

jaymzh: it wasn't an issue with upstream it was a limitation in how Actions which, which has been fixed.

test/run-broken.sh
```diff
@@ -15,7 +15,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -e
 set -o verbose
+bundle exec chef-zero &
+ZERO=$!
+trap "kill -9 $ZERO" EXIT
```
davide125, line 22: Do we really want to `-9` here? This probably warrants a comment either way (same for the two other scripts).

jaymzh: -9 seems right. we're exiting/being killed, we 100% want chef-zero to due immediately. Plus it's all in-memory so there's no danger.

NaomiReeves: I'm really just mirroring what we have in [grocery-delivery](https://github.com/facebook/grocery-delivery/blob/master/test/run-git.sh) for these tests.

test/run-broken.sh
```diff
@@ -15,14 +15,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+set -e
 set -o verbose
-rm -rf /tmp/ops
+bundle exec chef-zero &
+ZERO=$!
+trap "kill -9 $ZERO" EXIT
+SOURCE=/tmp/ops
```
davide125, line 23: This should probably use `mktemp -d` for safety

davide125 (approved):

.github/workflows/ci.yml
jaymzh: ugh, that's sad becasue it actually works now, but yeah kill it.

test/run-broken.sh
NaomiReeves: So the scripts that this calls (ie: `tt-auto.rb` and `tt-git.rb`) are referencing the specific repo `/tmp/ops` and while I can probably feed the temp dir to this, I'm not sure that I want to tackle that here too.

PR #161, opened 2021-04-14
Fix ARGV check, empty array is alway true

scottruss:
#157 added this check but its always truthy because it can be an empty array and some taste-tester commands (eg `test-taster upload`) fail the following check with the error message `Mode upload does not accept extra argument: []`


jaymzh (approved):
Gah. I missed that in my code review, sorry. Nice catch!

PR #162, opened 2021-04-16
Make sure the host is the last arg to SSH

jaymzh:
While what TT does today is perfectly acceptable to SSH, some users wrap
SSH, and those wrappers aren't always as smart as SSH, unfortunately. To
be as compatible as possible, leave the host arguement to the end

### DIFF
--- a/lib/taste_tester/ssh_util.rb
+++ b/lib/taste_tester/ssh_util.rb
@@ -69,7 +69,7 @@ module TasteTester
         else
           cmds = command_list.join(' && ')
         end
-        cmd = "#{ssh} #{ssh_target}"
+        cmd = "#{ssh} #{ssh_target} "
         cc = Base64.encode64(cmds).delete("\n")
         if TasteTester::Config.windows_target
 
--- a/lib/taste_tester/ssh_util.rb
+++ b/lib/taste_tester/ssh_util.rb
@@ -20,8 +20,11 @@ module TasteTester
           "-J #{TasteTester::Config.jumps}" : ''
         "#{TasteTester::Config.ssh_command} #{jumps} -T -o BatchMode=yes " +
           '-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ' +
-          "-o ConnectTimeout=#{TasteTester::Config.ssh_connect_timeout} " +
-          "#{TasteTester::Config.user}@#{@host} "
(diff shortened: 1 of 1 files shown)


jaymzh:
side note, this works only because the only users of `ssh_base_cmd` also use `build_ssh_cmd` - there's only two of each. If that wasn't the case, this would break things.

jaymzh:
Oh and PS, this is totally untested. @NaomiReeves hit this and I went digging since I was the last person to touch this code. So don't merge this until she has a chance to test it. :)

lib/taste_tester/ssh_util.rb
```diff
@@ -20,8 +20,11 @@ def ssh_base_cmd
           "-J #{TasteTester::Config.jumps}" : ''
         "#{TasteTester::Config.ssh_command} #{jumps} -T -o BatchMode=yes " +
           '-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ' +
-          "-o ConnectTimeout=#{TasteTester::Config.ssh_connect_timeout} " +
-          "#{TasteTester::Config.user}@#{@host} "
+          "-o ConnectTimeout=#{TasteTester::Config.ssh_connect_timeout}"
+      end
+
+      def ssh_target
+        "#{TasteTester::Config.user}@#{@host}"
```
NaomiReeves, line 27: we need to preserve the extra space after the host or this will fail because all the commands get appended in to it
```suggestion
        "#{TasteTester::Config.user}@#{@host} "
```

NaomiReeves (changes requested):
This works really well, just a minor change

lib/taste_tester/ssh_util.rb
jaymzh: doing that here seems a bit ... wrong. I'd rather do that on line 72.

NaomiReeves: I'm fine with either option

lib/taste_tester/ssh_util.rb
```diff
@@ -66,7 +69,7 @@ def build_ssh_cmd(ssh, command_list)
         else
           cmds = command_list.join(' && ')
         end
-        cmd = ssh
+        cmd = "#{ssh} #{ssh_target}"
```
NaomiReeves, line 72:
```suggestion
        cmd = "#{ssh} #{ssh_target} "
```

NaomiReeves (approved):

PR #170, opened 2021-07-26
Add base spec for taste-tester for ssh

nishchintraina:
Set up a baseline for testing in tt.
Add spec test for testing building of ssh command

### DIFF
--- a/spec/ssh_spec.rb
+++ b/spec/ssh_spec.rb
@@ -1,0 +1,108 @@
+# vim: syntax=ruby:expandtab:shiftwidth=2:softtabstop=2:tabstop=2
+
+# Copyright 2020-present Facebook
+#
+# Licensed under the Apache License, Version 2.0 (the 'License');
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+require 'logger'
+require 'base64'
+require 'between_meals/util'
+require_relative '../lib/taste_tester/logging'
+require_relative '../lib/taste_tester/exceptions'
+require_relative '../lib/taste_tester/ssh_util'
+require_relative '../lib/taste_tester/ssh'
+require_relative '../lib/taste_tester/config'
+
+describe TasteTester::SSH do
+  let(:logger) do
+    Logger.new('/dev/null')
+  end
+
+  let(:tt_ssh) do
+    TasteTester::SSH.new('mock_host')
+  end
+
+  context 'test default config' do
+    it 'test ssh base command' do
+      expect(tt_ssh.ssh_base_cmd).to eq(
+        'ssh  ' +
+        '-T ' +
+        '-o BatchMode=yes ' +
+        '-o UserKnownHostsFile=/dev/null ' +
+        '-o StrictHostKeyChecking=no ' +
+        '-o ConnectTimeout=5',
+      )
+    end
+
+    it 'test ssh target' do
+      expect(tt_ssh.ssh_target).to eq(
+        'root@mock_host',
+      )
+    end
+
+    it 'test build ssh command' do
+      expect(tt_ssh.build_ssh_cmd('mock_ssh', ['cmd1', 'cmd2'])).to eq(
+        'mock_ssh root@mock_host ' +
+        "\"echo 'Y21kMSAmJiBjbWQy' | base64 --decode | bash -x\"",
+      )
+    end
+  end
+
+  context 'test custom configs linux' do
+    before do
+      TasteTester::Config.ssh_connect_timeout 10
+      TasteTester::Config.jumps 'mock_jump_user@mock_jump_host'
+      TasteTester::Config.ssh_command 'mock_ssh_command'
+      TasteTester::Config.user 'rossi'
+    end
+
+    it 'test ssh base command' do
+      expect(tt_ssh.ssh_base_cmd).to eq(
+        'mock_ssh_command ' +
+        '-J mock_jump_user@mock_jump_host ' +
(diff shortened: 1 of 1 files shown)


nishchintraina:
@jaymzh @NaomiReeves - tunnel spec's coming in another PR :)

spec/ssh_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+# limitations under the License.
+
+require 'logger'
+require 'base64'
+require 'between_meals/util'
+require 'taste_tester/logging'
+require 'taste_tester/exceptions'
+require 'taste_tester/ssh_util'
+require 'taste_tester/ssh'
+require 'taste_tester/config'
```
jaymzh, line 24: you should change all these to:
```
require_relative '../lib/taste_tester/logging'
...
```

so the tests run against the repo and not installed code.

spec/ssh_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+describe TasteTester::SSH do
+  let(:logger) do
+    Logger.new('/dev/null')
+  end
+
+  tt_ssh = TasteTester::SSH.new('mock_host')
```
jaymzh, line 31:
```ruby
  let(:tt_ssh) do
    TasteTester::SSH.new('mock_host')
  end
```

spec/ssh_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    it 'test build ssh command' do
+      expect(tt_ssh.build_ssh_cmd('mock_ssh', ['cmd1', 'cmd2'])).to eq(
+        'mock_ssh root@mock_host ' \
+        "\"echo 'Y21kMSAmJiBjbWQy' | base64 --decode | bash -x\"",
+      )
+    end
+  end
+
+  context 'test custom configs linux' do
+    before(:each) do
```
jaymzh, line 60: why before _each_ ? Unless you're changing these in tests, you can just do `before do`

spec/ssh_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    it 'test ssh target' do
+      expect(tt_ssh.ssh_target).to eq(
+        'root@mock_host',
+      )
+    end
+
+    it 'test build ssh command' do
+      expect(tt_ssh.build_ssh_cmd('mock_ssh', ['cmd1', 'cmd2'])).to eq(
+        'mock_ssh root@mock_host ' \
```
jaymzh, line 53: backslashes are python-y. Use `+`

spec/ssh_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    it 'test ssh base command' do
+      expect(tt_ssh.ssh_base_cmd).to eq(
+        'mock_ssh_command ' \
+        '-J mock_jump_user@mock_jump_host ' \
+        '-T ' \
+        '-o BatchMode=yes ' \
+        '-o UserKnownHostsFile=/dev/null ' \
+        '-o StrictHostKeyChecking=no ' \
+        '-o ConnectTimeout=10',
```
jaymzh, line 75: Use + instead of \

jaymzh (changes requested):
This PR makes me *HAPPY*! Nice work.

Few minor things below, but that's it.

I was gonna say you have to actually modify the GH stuff so they run... but turns out they already do! Neat.

nishchintraina:
> This PR makes me _HAPPY_! Nice work.
> 
> Few minor things below, but that's it.
> 
> I was gonna say you have to actually modify the GH stuff so they run... but turns out they already do! Neat.

Thanks @jaymzh! I will follow up with the suggested fixes in a bit. <3

n00n question - I was trying to also test the exception message for failures, but the `error!` block inside `SSH:Util` calls a fail - and it doesn't get caught by the following
```
    it 'test ssh exception message' do
      expect(tt_ssh.error!).to raise_error(
        TasteTester::Exceptions::SshError
      )
    end
```

Any ideas?

jaymzh:
> n00b question - I was trying to also test the exception message for failures, but the `error!` block inside `SSH:Util` calls a fail - and it doesn't get caught by the following
> 
> ```
>     it 'test ssh exception message' do
>       expect(tt_ssh.error!).to raise_error(
>         TasteTester::Exceptions::SshError
>       )
>     end
> ```
> 
> Any ideas?

Yeah - you can't use `()`, you need a code-block to catch the error.

```ruby
expect { tt_ssh.error! }.to raise_error(...)
```

nishchintraina:
>     it 'test ssh exception message' do
>       expect(tt_ssh.error!).to raise_error(
>         TasteTester::Exceptions::SshError
>       )
>     end

Thanks! I think it's looks good-y now. :)

jaymzh (approved):
I appreciate the fact that you added the crazy windows target stuff in there too.

jaymzh:
You need to fix your lint/tests before I can merge though.

nishchintraina:
Huh yeah I was looking into this - I think there is a problem with how I am overriding the tt config - it breaks the tunnel tests - both pass individually but not together. :(

spec/ssh_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+      )
+    end
+  end
+
+  context 'test custom configs linux' do
+    before do
+      TasteTester::Config.ssh_connect_timeout 10
+      TasteTester::Config.jumps 'mock_jump_user@mock_jump_host'
+      TasteTester::Config.ssh_command 'mock_ssh_command'
+      TasteTester::Config.user 'rossi'
```
nishchintraina, line 79: @jaymzh  This is mutating shit globally, so it pollutes other tests. Sorry :( Once I find a way to do this I will fix the other test too.

jaymzh: Oh that makes sense, it's a singleton.

You can save the values, and then use an `after` block to restore them.

nishchintraina: Thanks! I was able to figure it out using the save/restore context in `Mixlib::Config` :)

jaymzh:
Please don't merge... rebase.

nishchintraina:
> Please don't merge... rebase.

I... don't git it. :(

I mean I do, but this time, I didn't. 

Anyway, should be good now. :)

jaymzh (approved):
Yay tests!

PR #181, opened 2021-12-09
Add parallel host operations with opt-in concurrent upload and structured results

Babar:
## Summary

Add bounded parallel host operations for `test`, `untest`, `run`, and `keeptesting`, rebased onto current upstream `main`. Host concurrency defaults to the local CPU count and is configurable through `--parallel-host-requests` or `parallel_hosts`. Chef output is prefixed with the hostname.

Upload completes before host configuration by default. Users can explicitly overlap these phases with `--unsafe-parallel-upload` or `unsafe_parallel_upload true` in their config. In this mode:

- The local server is initialized before hosts are configured.
- Default-visible warnings explain that host progress does not mean Chef can run safely, including when hosts finish while upload is still pending.
- The final upload-complete notification lists the successfully configured hosts only after upload succeeds.
- Upload failure suppresses readiness, exits with code 20, and tells users to successfully retry upload or untest the affected hosts. It does not automatically roll back host configuration.

## Results and review fixes

- Distinct exit codes cover all-failed versus partially-failed operations and SSH, already-testing, other, or mixed failure categories: 10–13 for total failures, 14–17 for partial failures; 0 remains success and 1 covers usage/unexpected errors.
- Optional `--results-json PATH` / `results_json '/path/results.json'` writes an atomic JSON report with per-host results, upload status, process exit code, and `ready_to_run_chef`. Upload failure never marks a host ready, even if its configuration succeeded. Existing `impact --json` behavior is unchanged.
- Fix thread-limit joins, avoid duplicate error-hook calls, preserve the originating command in error hooks, and distinguish remote Chef exit statuses from internal failure categories.
- Address review style comments, explicitly subclass `::IO`, replace the stdout descriptor literal with `STDOUT.fileno`, and bump the version to `0.1.0`.
- Document configuration, exit-code changes, JSON schema, and recovery behavior in README.

## Validation

- 56 RSpec examples pass, including upstream reverse-tunnel coverage and regressions for concurrency limits, upload ordering/failure, all exit-code categories, and JSON reporting.
- RuboCop: no offenses. README Markdown lint passes.
- No-op CLI smoke tests verify JSON output via the CLI flag and config setting without contacting hosts.
- Local tests used the available Ruby 3.4.7 with a temporary validation Gemfile adding the `syslog` gem required by older Chef on Ruby 3.4. Ruby 2.6 syntax checks pass; the CI Ruby matrix has not been run locally.
- No live-host or internal-wrapper testing was performed.

Signed-off-by: Olivier Raginel <babar@fb.com>

### DIFF
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -409,6 +409,16 @@ MODES:
     ) do
       options[:skip_post_test_hook] = true
     end
+
+    opts.on(
+      '--parallel-host-requests NUMHOSTS', 'Start NUMHOSTS threads in parallel'
+    ) do |n|
+      unless n.to_i > 0
+        logger.error("Sorry, number of threads cannot be less than 1: #{n}")
+        exit(1)
+      end
+      options[:parallel_hosts] = n.to_i
+    end
   end
 
   if mode == 'help'
--- a/bin/taste-tester
+++ b/bin/taste-tester
@@ -411,6 +411,15 @@ MODES:
     end
 
     opts.on(
+      '-b', '--bundle BUNDLE_SETTING', ['true', 'false', 'compatible'], 'Bundle mode setting'
+    ) do |bundle_setting|
+      if ['true', 'false'].include?(bundle_setting)
+        bundle_setting = bundle_setting == 'true'
+      end
+      options[:bundle] = bundle_setting
+    end
+
+    opts.on(
       '--parallel-host-requests NUMHOSTS', 'Start NUMHOSTS threads in parallel'
     ) do |n|
       unless n.to_i > 0
--- a/lib/taste_tester/commands.rb
+++ b/lib/taste_tester/commands.rb
@@ -63,6 +63,69 @@ module TasteTester
       end
     end
 
+    def self.run_parallel(mode, hosts, repo = nil)
+      host_threads = []
+      tested_hosts = []
+      connect_failures = 0
+      server = TasteTester::Server.new
+      hosts.each do |hostname|
+        # Poor man thread pool manager: keeping it simple
+        nb_threads_over_max = host_threads.length - TasteTester::Config.parallel_hosts
+        if nb_threads_over_max >= 0
+          host_threads[nb_threads_over_max].join
+        end
+        host_threads << Thread.new do
+          Thread.current[:hostname] = hostname
+          Thread.current.report_on_exception = false
+          TasteTester::Host.new(hostname, server).send mode
+        end
+      end
+      host_threads.each do |host_thread|
+        begin
+          host_thread.join
+          hostname = host_thread[:hostname]
+          tested_hosts << hostname
+        rescue TasteTester::Exceptions::AlreadyTestingError => e
+          logger.error("User #{e.username} is already testing on #{hostname}")
+        rescue TasteTester::Exceptions::SshError
+          logger.error("Cannot connect to #{hostname}")
+          connect_failures += 1
+        rescue StandardError => e
+          # Call error handling hook and re-raise
+          TasteTester::Hooks.post_error(TasteTester::Config.dryrun, e,
+                                        __method__, hostname)
+          raise
+        end
+      end
+      if mode == :test
+        puts "Repo is: #{repo} -- parallel"
+        unless TasteTester::Config.skip_post_test_hook ||
+            TasteTester::Config.linkonly
+          TasteTester::Hooks.post_test(TasteTester::Config.dryrun, repo,
+                                      tested_hosts)
+        end
+      end
+      if tested_hosts.to_set == hosts.to_set
+        # No exceptions, complete success: every host listed is now configured
+        # to use our chef-zero instance.
+        exit(0)
+      end
+      if tested_hosts.empty?
+        if connect_failures > 0
+          # All the hosts we had failed, with at least one because of ssh
+          exit(1)
+        end
+        # All requested hosts are being tested by another user. We didn't change
+        # their configuration.
+        exit(3)
+      end
+      # Otherwise, we got a mix of success and failure due to being tested by
+      # another user. We'll be pessemistic and return an error because the
+      # intent to taste test the complete list was not successful.
+      # code.
+      exit(2)
+    end
+
     def self.test
       hosts = TasteTester::Config.servers
       unless hosts
@@ -82,7 +145,6 @@ module TasteTester
         end
         upload
       end
-      server = TasteTester::Server.new
       unless TasteTester::Config.linkonly
         if TasteTester::Config.no_repo
           repo = nil
@@ -101,44 +163,8 @@ module TasteTester
           TasteTester::Config.linkonly
         TasteTester::Hooks.pre_test(TasteTester::Config.dryrun, repo, hosts)
       end
-      tested_hosts = []
-      hosts.each do |hostname|
-        host = TasteTester::Host.new(hostname, server)
-        begin
-          host.test
-          tested_hosts << hostname
-        rescue TasteTester::Exceptions::AlreadyTestingError => e
-          logger.error("User #{e.username} is already testing on #{hostname}")
-        rescue StandardError => e
-          # Call error handling hook and re-raise
-          TasteTester::Hooks.post_error(TasteTester::Config.dryrun, e,
-                                        __method__, hostname)
-          raise
-        end
-      end
-      unless TasteTester::Config.skip_post_test_hook ||
-          TasteTester::Config.linkonly
-        TasteTester::Hooks.post_test(TasteTester::Config.dryrun, repo,
-                                     tested_hosts)
-      end
-      # Strictly: hosts and tested_hosts should be sets to eliminate variance in
-      # order or duplicates. The exact comparison works here because we're
-      # building tested_hosts from hosts directly.
-      if tested_hosts == hosts
-        # No exceptions, complete success: every host listed is now configured
-        # to use our chef-zero instance.
-        exit(0)
-      end
-      if tested_hosts.empty?
-        # All requested hosts are being tested by another user. We didn't change
-        # their configuration.
-        exit(3)
-      end
-      # Otherwise, we got a mix of success and failure due to being tested by
-      # another user. We'll be pessemistic and return an error because the
-      # intent to taste test the complete list was not successful.
-      # code.
-      exit(2)
+      puts "Repo is: #{repo} -- before parallel"
+      self.run_parallel(:test, hosts, repo)
     end
 
     def self.untest
@@ -147,18 +173,7 @@ module TasteTester
         logger.error('You must provide a hostname')
         exit(1)
       end
-      server = TasteTester::Server.new
-      hosts.each do |hostname|
-        host = TasteTester::Host.new(hostname, server)
-        begin
-          host.untest
-        rescue StandardError => e
-          # Call error handling hook and re-raise
-          TasteTester::Hooks.post_error(TasteTester::Config.dryrun, e,
-                                        __method__, hostname)
-          raise
-        end
-      end
+      self.run_parallel(:untest, hosts)
     end
 
     def self.runchef
@@ -167,18 +182,7 @@ module TasteTester
         logger.warn('You must provide a hostname')
         exit(1)
       end
-      server = TasteTester::Server.new
-      hosts.each do |hostname|
-        host = TasteTester::Host.new(hostname, server)
-        begin
-          host.runchef
-        rescue StandardError => e
-          # Call error handling hook and re-raise
-          TasteTester::Hooks.post_error(TasteTester::Config.dryrun, e,
-                                        __method__, hostname)
-          raise
-        end
-      end
+      self.run_parallel(:runchef, hosts)
     end
 
     def self.keeptesting
@@ -187,18 +191,7 @@ module TasteTester
         logger.warn('You must provide a hostname')
         exit(1)
       end
-      server = TasteTester::Server.new
-      hosts.each do |hostname|
-        host = TasteTester::Host.new(hostname, server)
-        begin
-          host.keeptesting
-        rescue StandardError => e
-          # Call error handling hook and re-raise
-          TasteTester::Hooks.post_error(TasteTester::Config.dryrun, e,
-                                        __method__, hostname)
-          raise
-        end
-      end
+      self.run_parallel(:keeptesting, hosts)
     end
 
     def self.upload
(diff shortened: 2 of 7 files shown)


jaymzh:
IIRC, this broke internal wrappers... so CC @NaomiReeves on that.

But you definitely should make sure you've tested this with internal wrappers.

Babar:
The other PR broke the wrappers, from what I remember. The part where I moved the ARGS to be hostnames, 'cause some parts of our wrappers rely on TT just dropping unknown parameters.
I've fixed the wrapper too, but this is mostly to get speed improvements when testing a bunch of hosts.
Also, I've fixed the lint, so updating the PR shortly.

Babar:
Oh one thing I'd love feedback on:
* How do people think we should deal with failures?
* How do we deal with logs?
In `test`, we report the failures and exit with:
* 0 if no failure
* 1 if all hosts failed and at least one failure was due to ssh connection issue
* 2 otherwise (meaning at least one host failed for whatever reason and at least one succeeded)
* 3 if all hosts failed because they were already tested by another user
I kept this logic for the other modes (so `untest`, `run` and `keeptesting`), even though 3 cannot happen.

As for logging, `run` can get very noisy with all the threads writing at the same time, and not knowing what's what. So I started writing something to prefix the logs with the hostname. But then I went down the rabbit hole, so I'll try to finish this some other night.

lib/taste_tester/commands.rb
```diff
@@ -63,6 +63,69 @@ def self.status
...
+    def self.run_parallel(mode, hosts, repo = nil)
+      host_threads = []
+      tested_hosts = []
+      connect_failures = 0
+      server = TasteTester::Server.new
+      hosts.each do |hostname|
+        # Poor man thread pool manager: keeping it simple
+        nb_threads_over_max = host_threads.length - TasteTester::Config.parallel_hosts
+        if nb_threads_over_max >= 0
+          host_threads[nb_threads_over_max].join
```
gogsbread, line 75: shouldn't this `join` be rescued as well?

jaymzh:
> Oh one thing I'd love feedback on:
> 
> * How do people think we should deal with failures?
> * How do we deal with logs?
>   In `test`, we report the failures and exit with:
> * 0 if no failure
> * 1 if all hosts failed and at least one failure was due to ssh connection issue
> * 2 otherwise (meaning at least one host failed for whatever reason and at least one succeeded)
> * 3 if all hosts failed because they were already tested by another user
>   I kept this logic for the other modes (so `untest`, `run` and `keeptesting`), even though 3 cannot happen.
> 
> As for logging, `run` can get very noisy with all the threads writing at the same time, and not knowing what's what. So I started writing something to prefix the logs with the hostname. But then I went down the rabbit hole, so I'll try to finish this some other night.

I'm glad you asked!

A lot of this is actually the kind of stuff that we (we = TM folks, not FB folks, so past-past-past-life) solved over in https://github.com/txcketmaster-xx/onall - it forks a bunch of ssh's, joins 'em all back, and handles output nicely. It has a few options - live, buffered, prefixed-by-hostname, in-a-directory one-file-per-host, etc. and I think it's a great model for how to manage output and make it usable for folks.

As for return values: 0 - no errors, we can all agree on that. But maybe some more nuance for errors:

* All failed, all SSH failures (no test-setup/run failures)
* All failed, all because in-testing
* All failed, other test-setup/run failures
* All failed, mix of issues
* Some failed, all SSH related
* Some failed, all because in-testing
* Some failed, other test-setup/run failures
* Some failed, mix of issues

Further, I might recommend outputing a JSON structure of host=>failure, perhaps on stderr, or perhaps to a configurable file.

Babar:
Stupid question, but what's the difference between `onall` and `parallel`? They're bot written in perl, and seem to be doing pretty much the same thing. I mean, `parallel --tag ssh {} ...`. Haven't read the entire code yet, but I know `parallel`'s code fairly well, which is why I'm asking.

lib/taste_tester/commands.rb
Babar: Good point. I guess this never happened in my testing 'cause I have a gazillion of cores :)

lib/taste_tester/host.rb
```diff
@@ -26,6 +26,29 @@
 require 'taste_tester/exceptions'
 
 module TasteTester
+  # Wrapper around IO to stream, prefixing the lines with the hostname
+  class IO < IO
```
dafyddcrosby: Makes this a little easier to understand
```suggestion
  class IO < ::IO
```

lib/taste_tester/commands.rb
```diff
@@ -63,7 +63,94 @@ def self.status
...
+        if print_remaining
+          host_list = long_running_threads.map { |t| t[:hostname] }.join ','
+          printf("Still waiting on #{host_list} to finish...")
+          print_remaining = false
+        end
+        long_running_threads.each do |host_thread|
+          thread_status = _join_and_handle_ssh_exception(host_thread, 5)
+          unless thread_status.nil?
+            long_running_threads.delete host_thread
+            return_code.merge! [thread_status].to_h
```
jaymzh:
```suggestion
            return_code.merge!([thread_status].to_h)
```

lib/taste_tester/commands.rb
```diff
@@ -63,7 +63,94 @@ def self.status
...
+      if threaded_upload
+        all_threads << Thread.new do
+          Thread.current[:hostname] = :upload
+          Thread.current.report_on_exception = true
+          Thread.current[:status] = upload
+        end
+        running_threads += 1
+      end
+
+      hosts.to_set.each do |hostname|
```
jaymzh: we can't start these until the upload is done.

lib/taste_tester/host.rb
jaymzh: David is right - this looks like your subclassing yourself, we should be clear.

lib/taste_tester/host.rb
```diff
@@ -50,16 +73,9 @@ def runchef
       transport = get_transport
       transport << TasteTester::Config.chef_client_command
 
-      io = IO.new(1)
+      io = TasteTester::IO.new(name, 1)
```
jaymzh: magic numbers are bad. What is 1?

jaymzh (changes requested):

lib/taste_tester/commands.rb
Babar: Why? There is no point in waiting for the upload. You can't run it, but setting a host to taste-test has almost no relation with upload. The only caveat is that you need to ensure you have a working chef-zero instance, so you know the port number. But upload itself can totally be done in parallel.

lib/taste_tester/commands.rb
```diff
@@ -80,9 +167,8 @@ def self.test
         if TasteTester::Config.linkonly
           logger.warn('Ignoring --linkonly because --really not set')
         end
-        upload
+        do_upload = true
```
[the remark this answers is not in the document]
Babar: I disagree. I initially did what you said, but then realized upload was taking longer and longer, so made a thread for it. Why do you think it has to complete before we set hosts to testing?

lib/taste_tester/host.rb
Babar: You do realize I didn't change that, right? I just added the name as a parameter :) I also have no idea what `1` means here.

taste_tester.gemspec
```diff
@@ -14,7 +14,7 @@
 
 Gem::Specification.new do |s|
   s.name = 'taste_tester'
-  s.version = '0.0.19'
+  s.version = '0.0.21'
```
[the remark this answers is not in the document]
Babar: Honestly, I just like 21 :) We totally should change that.

lib/taste_tester/commands.rb
```diff
@@ -63,7 +63,95 @@ def self.status
...
+    end
+
+    def self._join_and_handle_ssh_exception(host_thread, limit = 15)
+      hostname = host_thread[:hostname]
+      logger.info("Joining thread for #{hostname}...")
+      return if host_thread.join(limit).nil?
+      return hostname, host_thread[:status]
+    rescue TasteTester::Exceptions::AlreadyTestingError => e
+      logger.error("User #{e.username} is already testing on #{hostname}")
+      return hostname, 42
```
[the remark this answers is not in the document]
Babar: IIRC, I used the same constants as the current implementation. But I agree it's better to use actual constants.

lib/taste_tester/commands.rb
```diff
@@ -63,7 +63,94 @@ def self.status
...
+        if thread_status.nil?
+          # We timed out waiting for it to finish
+          long_running_threads << host_thread
+        else
+          return_code.merge! [thread_status].to_h
+        end
+      end
+
+      print_remaining = true
+      while !long_running_threads.empty?
```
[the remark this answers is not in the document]
Babar:
```suggestion
      until long_running_threads.empty?
```

lib/taste_tester/commands.rb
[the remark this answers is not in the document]
jaymzh: Because it's entirely possible someone sees web12345 is done in the output, and goes and runs Chef on it. But in fact, that may not be safe to do for a long time. Remember that we're not necessarily starting with a blank chef-zero where the chef run will break early. It could be an existing one which month-old configs and you could revert that machine (or machines!) to a completely non-working state. A pretty common workflow would be that I parallel test 10 hosts, and those 10 hosts are machines I"m already SSH'd to because I"m preping the test. As I see the names go by from `tt`, I run chef. You're now expecting people to have an unreasonable level of understanding of internals to know that's not safe.

I feel pretty strongly this is the wrong design.

Even for,  autotest or something like it where the automation understands and is going to wait for all threads anyway, I can see wanting to save some time here, so let's discuss that possibility. Let's say we really want this and we hide it behind a flag like --unsafe-parallel-upload and make the help very clear no hosts can have chef run until TT exits successfully.  If we do that - what happens when every host succeeds but the upload fails? How do we indicate that? Or, what about when some hosts succeed, some fail, and the upload fails. Now we report partial success... except, that's not true either, even the "successful" hosts cannot safely have Chef run on them (see aforementioned old chef-zero state). Worse, we're now in a super inconsistent state where a bunch of hosts are pointed do a Chef server with a completely unknown state that requires either strong guarantees about the parent system cleaning them up in a timely fashion, or manual intervention from a human post-haste.

All this comes down to the design principal in the current system we don't touch a production host until _after_ we're sure we're pointing it at what the user expects. We don't point it at unknown stuff and hope for the best, and this code does just that, which is incredibly dangerous.

lib/taste_tester/host.rb
jaymzh: I do, but you touch the line you get to fix it. :)

lib/taste_tester/commands.rb
jaymzh: See other longer comment

bin/taste-tester
```diff
@@ -74,13 +74,13 @@ MODES:
...
-        tested by other users.
-      3 All hosts were not taste-tested because they are being taste-tested by
-        other users.
+    Host-operation exit codes (also used by untest, run, and keeptesting):
+      0 Success; 1 usage or unexpected command error.
+      10 All failed: SSH; 11 all failed: already testing;
+      12 all failed: other errors; 13 all failed: mixed errors.
+      14 Partial failure: SSH; 15 partial failure: already testing;
+      16 partial failure: other errors; 17 partial failure: mixed errors.
+      20 Upload failed: no hosts are safe to run Chef on.
```
jaymzh, line 83: This is awful to read. Keep one line per code, please.

I'm guessing this AI slop.

bin/taste-tester
```diff
@@ -423,6 +423,29 @@ MODES:
       end
       options[:bundle] = bundle_setting
     end
+
+    opts.on('--results-json PATH', 'Write host results and upload status as JSON to PATH') do |path|
```
jaymzh: fix line length

lib/taste_tester/commands.rb
```diff
@@ -135,16 +138,42 @@ def self._join_and_handle_ssh_exception(host_thread, limit = 15)
...
+        upload { started << true }
+      rescue StandardError, SystemExit => e
+        Thread.current[:upload_error] = e
+      ensure
+        # Also release the caller if starting the server failed.
+        started << true
+      end
+      # The server must exist and its port be known before configuring hosts.
+      started.pop
+      _finish_parallel_upload(thread) if thread[:upload_error]
```
jaymzh, line 165: This seems wrong. This is what joins the thread. It does stiff even outside of upload_error.

lib/taste_tester/commands.rb
```diff
@@ -61,61 +63,62 @@ def self.status
...
         end
         all_threads << Thread.new do
-          Thread.current[:hostname] = hostname
           Thread.current.report_on_exception = false
-          Thread.current[:status] = TasteTester::Host.new(hostname, server).send mode
+          status = TasteTester::Host.new(hostname, server).send(mode)
+          Thread.current[:status] = [:test, :runchef].include?(mode) ? status : 0
         end
+        all_threads.last[:hostname] = hostname
+        all_threads.last[:mode] = mode
```
jaymzh, line 96: The previous code of `Thread.current[:hostname] = hostname` inside the block seemed way better.

jaymzh (changes requested):

lib/taste_tester/commands.rb
```diff
@@ -61,7 +65,186 @@ def self.status
...
+          Thread.current.report_on_exception = false
+          status = TasteTester::Host.new(hostname, server).send(mode)
+          Thread.current[:status] = [:test, :runchef].include?(mode) ? status : 0
+        end
+        all_threads.last[:hostname] = hostname
+        all_threads.last[:mode] = mode
+        running_threads += 1
+      end
+
+      all_threads.drop(last_running_thread).each do |host_thread|
```
jaymzh, line 167: drop doesn't returned dropped things:
```
irb(main):006> x = ['a', 'b']
=> ["a", "b"]
irb(main):007> x.drop(2).each do |x| puts x end
=> []
irb(main):008> 
```

it's not clear this code has been tried

jaymzh (changes requested):
In general, I'm now OK with the approach. It's guarded with an unsafe option, there's lots of warnings, and some good guardrails. 

However, the code quality seems ot have dropped a lot since the last round - I'm guessing AI Slop is responsible. Please review and try the output of AI gen.

See review, above.

lib/taste_tester/commands.rb
jaymzh: This is still totally broken...

PR #182, opened 2023-03-03
Fix track_symlinks reference in taste-tester impact

dtheyer:
[@track_symlinks is set for TasteTester::Client](https://github.com/dtheyer/taste-tester/blob/main/lib/taste_tester/client.rb#L67), but not for TasteTester::Commands so this reference is always interpreted as false. The impact of this is that taste-tester impact is unable to detect changed files that are symlinked into the TasteTester::Config.relative_cookbook_dirs.

Since we are already referencing other config options within this function we can just read TasteTester::Config.track_symlinks directly.

Tested locally and verified that when track_symlinks is specified as true in the config, taste-tester impact adds the files to the returned changeset


jaymzh (approved):
This feels like an obvious bugfix.

PR #185, opened 2023-08-02
[github] drop Ruby 2.4 from CI, add 2.7

dafyddcrosby:


jaymzh (changes requested):
Fine in theory, but as with the BM PRs, lets get the tests sorted.

jaymzh (approved):

PR #199, opened 2026-05-22
Add run-reverse-tunnels command

cooperlees:
Sometimes we have servers in a firewalled off part of the network and we need to test some chef changes during server provisioning. The test config for the chef client is live and it's awaiting just a reverse tunnel to appear to start the chef run.

Due to this I would love a wrapper to start the tunnel(s) rather than manual `ssh -R [remote_port]:[local_host]:[local_port] [user]@[remote_host]` and manually looking at what port the chef instance is running on.


jaymzh (commented):
Uh...  this already exists. Just pass in `--use-ssh-tunnels`. Am I missing something here? Like the only concern here is that the config is already there, so you're worried about... the extra work to write it again?

cooperlees:
More so to skip writing incorrect config due to $reasons + much fun. Think secure clusters but worse ...

spec/commands_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+  before do
+    TasteTester::Config.restore(config_hash)
+  end
+
+  after do
+    TasteTester::Config.restore(config_hash)
+  end
+
+  describe '.run_reverse_tunnels' do
```
jaymzh: Nomenclature here is `#`, not `.`.

lib/taste_tester/commands.rb
```diff
@@ -199,6 +199,58 @@ def self.keeptesting
...
+      unless hosts
+        logger.warn('You must provide a hostname')
+        exit(1)
+      end
+
+      server = TasteTester::Server.new
+      unless TasteTester::Server.running? && server.port
+        logger.error('Local taste-tester server not running')
+        exit(1)
+      end
```
jaymzh: Why not start/upload if it's not running?

bin/taste-tester
```diff
@@ -112,6 +112,11 @@ MODES:
     NOTE!! This is #{'NOT'.red} a sufficient test, you must log onto the remote
     machine and verify the changes you are trying to make are actually present.
 
+  run-reverse-tunnels
+    Set up reverse SSH tunnels from each server in '-s' to your locally running
+    taste-tester server and wait until terminated. On shutdown (Ctrl+C, SIGTERM,
+    etc.), all tunnels are cleaned up.
```
jaymzh: Depending on below, we should note here that a server is required to be started and running.

jaymzh (changes requested):

jaymzh (approved):
Nice, LGTM! Thanks for addressing the feedback.

jaymzh:
Oh - there's some lint issues in the MDL. You didn't cause them but it'l break CI if you merge as-is, so please fix those long lines. Thanks.

dafyddcrosby (approved):

PR #200, opened 2026-06-21
Fix default databag dir

jaymzh:
This was a fun one.

Chef, when spinning up chef-zero, expects databags to be in 'data_bags',
but the default for tt is 'databags', so after testing stuff, it'll
then break production when there's no databags.

Signed-off-by: Phil Dibowitz <phil@ipom.com>


dafyddcrosby:
This unfortunate typo appears to be in use internally. Will need to figure out a graceful way to phase out that usage first 😞

jaymzh:
> This unfortunate typo appears to be in use internally. Will need to figure out a graceful way to phase out that usage first 😞 

Just add the typo to your config first so that it's explicit

